PR #2, opened 2016-07-27
Remove all use of eval

sj26:
Let's remove all use of eval from here!

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -52,7 +52,7 @@ function docker_compose_container_name() {
 
 # Runs the docker-compose command, scoped to the project, with the given arguments
 function run_docker_compose() {
-  command=(docker-compose)
+  local command=(docker-compose)
 
   # Append docker compose file
   command+=(-f "$(docker_compose_config_file)")
--- a/hooks/command
+++ b/hooks/command
@@ -6,43 +6,64 @@ set -ueo pipefail
 # environment variables into the desired BUILDKITE_PLUGIN_DOCKER_COMPOSE_
 # variables. This can be removed once we update the bootstrap for this new
 # plugin naming convention.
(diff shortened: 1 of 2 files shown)


hooks/command
```diff
@@ -6,43 +6,64 @@ set -ueo pipefail
...
 
-docker_compose_config_file() {
-  echo "${BUILDKITE_PLUGIN_DOCKER_COMPOSE_CONFIG:-docker-compose.yml}"
-}
-
 # Runs the docker-compose command, scoped to the project, with the given arguments
-run_docker_compose() {
-  buildkite-run "docker-compose -f $(docker_compose_config_file) -p $(docker_compose_project_name) $1"
+function run_docker_compose() {
+  command=(docker-compose)
```
toolmantim: Was this meant to be global? Needs a `local`

toolmantim:
I'd be tempted to simply remove all the `buildkite-prompt` and `buildkite-run` stuff and just run the commands w/o echoing them?

sj26:
It's really useful to see the commands when trying to debug. At the very least I think they should be output when BUILDKITE_DEBUG is true, but they're output within a squiggly section which gets hidden by default anyway.

hooks/command
sj26: Truth!

sj26: Good catch! Fixed.

toolmantim:
Nice nice. I wonder if we can provide these things to plugins? Or just provide it as a copy+paste helper that people add to their plugins?

toolmantim:
👍 🚀

sj26:
I don't really think we should export the `buildkite-run` etc... I'm tempted to rename them here.

I think it's up to plugin implementors to do everything beyond a basic process interface. They can definitely copy/paste useful bits! But I don't want to encourage utility libraries as I think that will quickly become a headache for all involved. :-)

PR #3, opened 2016-07-27
Multiple docker compose file

sj26:
Based on #2, and a port of buildkite/agent#357, let's enable the use of multiple docker compose files separated by colons.

We should probably also figure out as a next step after this is merged how to use Buildkite plugin config arrays as input, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,19 @@ steps:
         config: docker-compose.tests.yml
 ```
 
+or multiple config files:
+
+```yml
+steps:
+  - command: test.sh
+    plugins:
+      docker-compose:
+        run: app
+        config:
+          - docker-compose.yml
+          - docker-compose.test.yml
+```
+
 # Pre-building the image
 
 To speed up run parallel steps you can add a pre-building step to your pipeline, allowing all the `run` steps to skip image building:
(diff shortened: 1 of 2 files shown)


toolmantim:
Should we just go straight to supporting the array syntax, or are you thinking there should be two ways of setting multiple configs?

```yaml
plugins:
  docker-compose:
    config: docker-compose.yml
```

```yaml
plugins:
  docker-compose:
    config:
      - docker-compose.yml
      - docker-compose.ci.yml
```

as well as:

```yaml
plugins:
  docker-compose:
    config: "docker-compose.yml:docker-compose.ci.yml"
```

I'd always opt for just one nicely documented way of doing it, but if you think it could be super handy, and it'd still be clear in the Readme how to do it, we could support both!

sj26:
I think given we've just added support for this to the stable agent that we should probably retain forward compatibility for the moment. But we should probably add the array syntax support as well, in another PR.

toolmantim:
There was never any compatibility designed with this plugin though, it was meant for you to port. For example, there's no support for `BUILDKITE_DOCKER_COMPOSE_CONTAINER` which is the main one.

What about we go with arrays (I can add that now) and add we a section to the readme for "Migrating for Buildkite Agent 2.1 Docker Support" showing how to change it.

sj26:
I'm happy with whichever you prefer :-)

The migration path is pretty easy if you just have to split up your existing config into some more lines, so maybe skip this one and go straight to arrays if you think best!

toolmantim:
Sorry I meant we should probably do a migration section anyway, just to help people move from 2.0 to this plugin! So we can totally include instructions for that in there too.

sj26:
@toolmantim I've freshened this up and added support for arrays of config files, and updated the README a little. What do you think?

PR #7, opened 2016-09-26
Add a short sha tag to pushed images

anoldguy:
This is a proof-of-concept PR.

I love this plugin!!  This plugin made it a breeze to split up our tests and dropped our previous pipeline times from 10+ min down to under 5 for the worst case, and down to 2m30s in the best case!

However, the tags it produces, are, well...ugly.  They're completely cromulent on the backend, but we use those tag names in a deploy tool, and well, who wants to deploy `my-application-app-build-100`?  We're already used to the git short sha's, so I think it'd be nice to have an option to push that tag to the repo.

This commit adds the short sha (`git rev-parse --short $BUILDKITE_COMMIT`) as a tag that's pushed to the repo.

I'm not emotionally invested in this exact code change, so as long as we've got some way to indicate either "add the short sha as a tag" or "use this as a tag for this build", my request would be met.

Again, thanks for this great plugin!

### DIFF
--- a/hooks/commands/build.sh
+++ b/hooks/commands/build.sh
@@ -16,10 +16,14 @@ image_file_name() {
 
 push_image_to_docker_repository() {
   local tag="$DOCKER_IMAGE_REPOSITORY:$(image_file_name)"
+  local sha_tag="$DOCKER_IMAGE_REPOSITORY:$(short_sha)"
 
   plugin_prompt_and_must_run docker tag "$COMPOSE_SERVICE_DOCKER_IMAGE_NAME" "$tag"
+  plugin_prompt_and_must_run docker tag "$COMPOSE_SERVICE_DOCKER_IMAGE_NAME" "$sha_tag"
   plugin_prompt_and_must_run docker push "$tag"
+  plugin_prompt_and_must_run docker push "$sha_tag"
   plugin_prompt_and_must_run docker rmi "$tag"
+  plugin_prompt_and_must_run docker rmi "$sha_tag"
 
   plugin_prompt_and_must_run buildkite-agent meta-data set "$(build_meta_data_image_tag_key "$COMPOSE_SERVICE_NAME")" "$tag"
 }
--- a/hooks/command
+++ b/hooks/command
@@ -67,6 +67,10 @@ function build_meta_data_image_tag_key() {
   echo "docker-compose-plugin-built-image-tag-$1"
 }
 
+function short_sha(){
+  git rev-parse --short "$BUILDKITE_COMMIT"
+}
+
 ## BUILD OR RUN
 
 DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"


sj26:
Thanks so much for the contribution and the kind works, @anoldguy!

I love this idea. I'd love to give folks the ability to opt in to tagging their docker images by commit (full and abbreviated), and branch, and maybe some other things — build? job? We could make these plugin options which folks can opt in to. Maybe pushing the weird docker compose tag could be opted out, too, although it's important to maintain the ability to trace the lineage of a pushed imaged. But just wrapping your current PR in a config option would be a great start, and I'm happy to add this. Maybe aiming for a syntax like:

```yaml
steps:
  - command: test.sh
    plugins:
      docker-compose:
        build: app
        image-repository: index.docker.io/example/application
        tags:
          - "literal tag"
          # Variables work on the beta agent, so this should work
          - $BUILDKITE_COMMIT
          - $BUILDKITE_COMMIT_ABBREV # this one doesn't exist yet, though
          - ${BUILDKITE_COMMIT:0:7} # or maybe we add more parameter expansion options?
          - $BUILDKITE_BRANCH
```

The only blocker here is if there are multiple agents on the same machine sharing a docker daemon which end up building the same commit then we'll have a race condition around which job gets to tag and push which image. The remote moving is not a problem, but the local tag could change between `docker tag` and `docker push`, or be removed by a job while another is trying to push it.

I'm not sure how to solve this. I'd love a `docker push [repo] [image id]:[remote label]` like `git push [repo] [commit]:[remote ref]`, but I can't find an equivalent. Do you have any thoughts?

iroller:
Hi there. I wanted to add that docker-compose [has](https://docs.docker.com/compose/compose-file/) the following option:
```
If you specify image as well as build, then Compose names the built image with the webapp and optional tag specified in image:

build: ./dir
image: webapp:tag
```

It'd be great to support this method and push `webapp:tag` to docker registry.

Also you can use interpolation and do something like this in docker-compose.yml:
```
image: company/webapp:${BUILDKITE_COMMIT}
```
_(I tried it, it works)_

The only missing part here is for docker-compose-buildkite-plugin to push the proper tag. Adding a new tag isn't even required unless you need multiple tags.

anoldguy:
Hi @sj26!

Waitwaitwaitwaitwait. You can use ENV vars in the pipeline steps yml?!?

\<mind_blown.gif\>

Ok, if you could use your proposed syntax, that'd be awesome!  However, I don't have an answer for the docker race condition you mention, so... ¯\\\_(ツ)_/¯   We're pretty flexible in the implementation, tho, so let's not rule anything out yet.

You know, I just noticed that the plugin uses $BUILDKITE_JOB_ID (sans `-`) to tag the existing images in `COMPOSE_SERVICE_DOCKER_IMAGE_NAME`.  Until we get a better handle on how this could work for more customers than just us, I'll write a step that calculates the equivalent of `COMPOSE_SERVICE_DOCKER_IMAGE_NAME` and tags/pushes after all the tests have completed. It's not perfect, but it works within the framework without too many contortions.

I don't suppose we could expose the `COMPOSE_SERVICE_DOCKER_IMAGE_NAME` as a short-term workaround?

My end goal is to have a tag in my image repo that I can specify. Any solution you can provide and support easily is better than a perfect one that's difficult to maintain or would exist just for us.

jufemaiz:
In terms of requirements and outcome, how does this compare with https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/8 ?

hooks/commands/build.sh
```diff
@@ -16,10 +16,14 @@ image_file_name() {
 
 push_image_to_docker_repository() {
   local tag="$DOCKER_IMAGE_REPOSITORY:$(image_file_name)"
+  local sha_tag="$DOCKER_IMAGE_REPOSITORY:$(short_sha)"
 
   plugin_prompt_and_must_run docker tag "$COMPOSE_SERVICE_DOCKER_IMAGE_NAME" "$tag"
+  plugin_prompt_and_must_run docker tag "$COMPOSE_SERVICE_DOCKER_IMAGE_NAME" "$sha_tag"
   plugin_prompt_and_must_run docker push "$tag"
+  plugin_prompt_and_must_run docker push "$sha_tag"
```
jufemaiz: Is this pushing multiple images?

anoldguy: Yes, mainly because I didn't have a good handle on how to provide any tag thru from the config like @sj26 proposed.

If anything, it's a starter for a conversation, not a production ready PR. 👍🏻

As far as its relation to #8, I see they share a desire for more control of the pushed image tag, but we have no need for chained docker-compose.yml files.

jufemaiz: No worries, we were looking at it and just found it a tad odd.

Agreed on the pushed image tag. We've looked at the need for chained docker-compose.yml and seen a use case, hence @L226's PR.

sj26:
Sorry we haven't moved on this one yet. I've just proposed substring expansion over on the agent in https://github.com/buildkite/agent/pull/387. I'll see if we can add a config array for tags here, too. Tag collisions might just be a caveat of the feature. :-)

sj26:
@anoldguy I've taken an stab at the config syntax I posted earlier to enable specifying custom tags over in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/compare/custom-tags, any chance you could take it for a drive and see how it feels?

To achieve your short sha tag with your current agents you could switch to a [dynamic pipeline](https://github.com/buildkite/agent-tests/blob/master/tests/basic-pipeline/.buildkite/pipeline.sh) and interpolate into your pipeline.yml. I've also release a new agent beta, [v3.0.beta-11](https://github.com/buildkite/agent/releases/tag/v3.0-beta.11), which contains substring expansion (`${BUILDKITE_COMMIT:0:7}`)!

anoldguy:
We're running the beta stack of the elastic-ci-stack-for-aws, but that's only picking up beta-9 atm. I'll see about trying to get a new agent on some of them to test!

sj26:
Sorry @anoldguy, I think we forgot to rebuild the elastic-ci-stack after the most recent beta release! I've pusher another build which should have updated the cloudformation with a new ami here:

https://s3.amazonaws.com/buildkite-aws-stack/aws-stack.json

anoldguy:
Ok, so after re-reading this, perusing #9, #22, and #28, I see I can't specify a custom tag using the master branch.  Back in October, I [created a fork and branch](https://github.com/basecamp/docker-compose-buildkite-plugin/commits/fix-docker-compose-rewriting) to work on this code and I've got a version that's been working for us since then, but the changes were fairly drastic, and haven't aged well with the changes since then to the main master branch.

Is there any chance of this feature (regardless of implementation) getting added?

If docker supported deleting tags via the API, this wouldn't be a problem, I'd write a custom step to push the custom tag after tests have run, and then untag the BK generated names.

I don't mind writing the feature again, against master, but I'd like to make sure that I'm not working on something that people don't think belongs in the plugin.

❤️ @anoldguy

lox:
Hrm, My understanding was that #28 implemented a `[image-name](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#image-name-optional)` config option that lets you choose an image name:

```
steps:
  - name: ":docker: Build"
    agents:
      queue: docker-builder
    plugins:
      docker-compose:
        build: app
        image-name: ${BUILDKITE_PIPELINE_SLUG}-${BUILDKITE_COMMIT}
        image-repository: index.docker.io/org/repo
```

I'm slightly confused because I think that `image-name` in this context provides a tag, not an imagename, but I think it should do what you want @anoldguy? If not, it's a bug and we will fix it.

anoldguy:
Thanks for the reply, @lox! I'll give that a try. Sorry, I completely missed that image-name was standing in for tag. 😳

Regardless, I'll close this PR, as it's really not relevant anymore! (We should probably do the same for #9, right?)

sj26:
@anoldguy thanks for the reminder! I've updated and closed #9.

Issue #12, opened 2016-11-11
Docker-compose service with image and build breaks run


lox:
docker-compose recently introduced the ability to declare an `image` and a `build`:

https://docs.docker.com/compose/compose-file/#/image

> If the image does not exist, Compose attempts to pull it, unless you have also specified build, in which case it builds it using the specified options and tags it with the specified tag.

So for an example like this:

```
  accounts: 
    image: my-image/llamas:latest
    build:
      context: .
      dockerfile: ops/Dockerfile
    ports:
      - "9005:9005"
    depends_on:
      - mysql
```

This doesn't work with https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/hooks/commands/build.sh#L34.

PR #13, opened 2016-11-11
Use a configuration file to load the pre-built image

ianyamey:
Adds support for docker-compose services that have complex (multiline) `build` configurations in their yaml files.

**Context**
A docker-compose.yml allows the `build` configuration to be a string:

```
app:
  build: .
```

or an object

```
app:
  build:
    context: .
    dockerfile: Dockerfile-alternative
    args:
     somearg: foo
```

**Solution**
Instead of using `sed` to replace the `build: .` line with a path to the prebuilt image, the plugin will generate a docker-compose.buildkite-app-override.yml file of the form:

```
version: '2'
services:
  app:
    image: your-docker-registry.io/path-to/prebuilt-image
```

When the `docker-compose run` command is called, it first checks to see if the pre-built image exists before building it.

**Open questions**
- [x] Are there any users of this plugin that use the version 1 syntax for their docker-compose files?
  - If so, we can test for the presence of `version: '2'` in their docker compose file, and revert to the `sed` replacement for v1 users. For simplicity, I opted to leave this out until I hear that it's a requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ The default is `""`  which only builds images on the local Docker host doing the
 
 Note: this option only needs to be specified on the build step, and will be automatically picked up by following steps.
 
+Only works with version '2' docker-compose.yml configuration files
+
 This option can also be configured on the agent machine using the environment variable `BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_REPOSITORY`.
 
 ## Roadmap
--- a/hooks/commands/run.sh
+++ b/hooks/commands/run.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 COMPOSE_SERVICE_NAME="$BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN"
+COMPOSE_SERVICE_OVERRIDE_FILE="docker-compose.buildkite-$COMPOSE_SERVICE_NAME-override.yml"
 
 check_required_args() {
   if [[ -z "${BUILDKITE_COMMAND:-}" ]]; then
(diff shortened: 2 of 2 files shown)


lox:
Just ran into this issue.

lox:
Thoughts @sj26 / @toolmantim? Other option would be to simply tag the pulled image correctly so that it is picked up by docker-compose. Potentially racey though.

sj26:
I like this approach much better!

@lox: Have you tried it? Does it work for you?

sj26:
(I'm not sure we need to support version 1 configuration. Seems like that's pretty much died out.)

lox:
Just trying it now.

lox:
Seems to work well. 👍 from me.

lox:
This technique would also work nicely for the build phase to solve #12.

toolmantim:
Ah, this was on the todo list for this plugin… though I was thinking we'd make use of the changes in #3 — but this is simpler and nicer! Awesome.

I'm happy to require v2 of the docker-compose config file. Does Docker Compose give you a nice error message? If it does, we don't really have to have our own check… the Docker Compose error would be enough.

lox:
It will just not do any of the caching stuff with v1.

toolmantim:
> It will just not do any of the caching stuff with v1

So it simply ignores the extra (first) config file, and doesn't throw an error?

lox:
Probably merits some testing.

lox:
Actually, on re-read, is there any reason to think docker-compose doesn't allow you to use a v2 override for a v1 file @ianyamey?

ianyamey:
@lox: I just assumed that it wouldn't support a mixed set of configuration. I didn't see anything explicitly documented, so it may work!

lox:
Yup, you are right: https://gist.github.com/lox/fb563b5e501e53393d6680ff5efd372f

> ERROR: Version mismatch: file ./docker-compose.yml specifies version 1 but extension file ./docker-compose.buildkite.yml uses version 2.0

IMO that's an ok error message. I'm going to merge this.

lox:
Thanks so much for the PR @ianyamey!

PR #14, opened 2016-11-13
Override image name in build

lox:
Following on from https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/13, this uses an override file for the `build` step with the generated `image` name, which docker-compose then uses as the tag name. 

This makes it easier to figuring out which images have been built, and also supports docker-compose files with existing image stanza's (see https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/12).

### DIFF
--- a/hooks/commands/build.sh
+++ b/hooks/commands/build.sh
@@ -4,7 +4,6 @@ COMPOSE_SERVICE_NAME="$BUILDKITE_PLUGIN_DOCKER_COMPOSE_BUILD"
 COMPOSE_SERVICE_DOCKER_IMAGE_NAME="$(docker_compose_container_name "$COMPOSE_SERVICE_NAME")"
 DOCKER_IMAGE_REPOSITORY="${BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_REPOSITORY:-}"
 COMPOSE_SERVICE_OVERRIDE_FILE="docker-compose.buildkite-$COMPOSE_SERVICE_NAME-override.yml"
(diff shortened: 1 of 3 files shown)


toolmantim:
This is great!

lox:
This is now working, as far as I can tell. Can you think of any issues with this approach @toolmantim?

hooks/commands/build.sh
```diff
@@ -15,23 +16,30 @@ image_file_name() {
 }
 
 push_image_to_docker_repository() {
-  local tag="$DOCKER_IMAGE_REPOSITORY:$(image_file_name)"
+  local tag="$BUILDKITE_PLUGIN_DOCKER_COMPOSE_BUILD_TAG"
 
-  plugin_prompt_and_must_run docker tag "$COMPOSE_SERVICE_DOCKER_IMAGE_NAME" "$tag"
   plugin_prompt_and_must_run docker push "$tag"
   plugin_prompt_and_must_run docker rmi "$tag"
```
lox: Should we be removing the image? I think probably not with the new approach?

toolmantim:
Just the question in #13, about "What exactly does this do in v1 of docker-compose.yml syntax" and if it's "Shows some sensible error and the user would know what to do" then ace bananas

hooks/commands/build.sh
toolmantim: Hrmm, not sure. What's your thinking there with probably not?

lox:
docker-compose shows an error about mixing v1 and v2.

hooks/commands/build.sh
lox: Well, if the user has provided an image name that is going to trump the generated tag name, it probably doesn't make sense to delete it does it? Otherwise it would have a cache miss the next time around.

PR #16, opened 2016-11-18
Add a Buildkite Pipeline for some basic testing

lox:
This runs through some very basic tests of the compose plugin.


toolmantim:
I'm going to merge this in, which includes the bug fix for running commands without a previous build step.

PR #17, opened 2016-11-22
Override image name in build

lox:
Following on from https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/13, this uses an override file for the `build` step with the generated `image` name, which docker-compose then uses as the tag name. 

This makes it easier to figuring out which images have been built, and also supports docker-compose files with existing image stanza's (see https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/12).


lox:
@toolmantim I closed #14 in favour of this one.

toolmantim:
Nice! Any readme updates to go along with this?

lox:
I don't think this will change anything in terms of the user behaviour, basically it just ensures that images are named consistently (see #12).

PR #9, opened 2016-10-06
Custom tags

sj26:
Rebound of #7 to add custom tags in a generally configurable way!
- [ ] Make sure split build/run pipelines have a known tag to download -- maybe the original tag still needs to be always pushed

PR #22, opened 2016-11-29
Feature/custom image tags

sjauld:
A rehash of #9 that just overrides the default tag - no fancy stuff for adding multiple tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ steps:
       docker-compose:
         build: app
         image-repository: index.docker.io/org/repo
-    
+
   - wait
 
   - name: ":docker: Test %n"
@@ -116,6 +116,18 @@ Note: this option only needs to be specified on the build step, and will be auto
 Only works with version '2' docker-compose.yml configuration files
 
 This option can also be configured on the agent machine using the environment variable `BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_REPOSITORY`.
+
+## `image-tag` (optional)
+
+If you want to use a custom tag, please set this option.
+
+The default is `"$project_name-$COMPOSE_SERVICE_NAME-build-$BUILDKITE_BUILD_NUMBER"`.
+
+Note: this option only needs to be specified on the build step, and will be automatically picked up by following steps.
+
+Only works with version '2' docker-compose.yml configuration files
+
+This option can also be configured on the agent machine using the environment variable `BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_TAG`.
 
 ## Roadmap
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -117,6 +117,18 @@ Only works with version '2' docker-compose.yml configuration files
 
 This option can also be configured on the agent machine using the environment variable `BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_REPOSITORY`.
 
+## `image-tag` (optional)
+
+If you want to use a custom tag, please set this option.
+
+The default is `"$project_name-$COMPOSE_SERVICE_NAME-build-$BUILDKITE_BUILD_NUMBER"`.
```
lox: Might be sensible to represent this with: `${project_name}-${COMPOSE_SERVICE_NAME}-build-${BUILDKITE_BUILD_NUMBER}`. Leads to less potential bash foot-guns :)

lox: Oh, I just realized the intention wasn't to have this value interpolated.

lox: We could probably support interpolation with [`envsubst`](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html).

jufemaiz: @lox any further thoughts on this? We're quite keen to see it adopted rather than continuing the monkey patch on our side :)

hooks/command
```diff
@@ -83,6 +83,11 @@ function run_docker_compose() {
     command+=(-f "docker-compose.yml")
   fi
 
+  # Append docker compose exconf file
```
lox: I'm confused, isn't this already supported by using multiple config files https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#example ?

lox:
@sjauld could I trouble you to flatten all those commits into a single one and rebase from master?

sjauld:
@lox Will do

sjauld:
Seems like a couple of strange commits got mixed in, I'll remove them

sjauld:
@lox should be good to go now!

sjauld:
@jufemaiz STOP MESSING MY BRANCH!

lox:
Looks good to me, @toolmantim?

jufemaiz:
@toolmantim what is the check that hasn't completed yet? :)

toolmantim:
Thanks for the contribution @sjauld, and sorry for the slow review.

I've just poured over the code and looked at the original `build.sh`. It turns out the `BUILDKITE_PIPELINE_SLUG` doesn't contain the project slug (it did when it was `BUILDBOX_PIPELINE_SLUG`), so the whole `image_file_name` function is not needed. Which means a bunch of stuff could be removed, which I've done in #28 

I also checked out the [docker tag docs](https://docs.docker.com/engine/reference/commandline/tag/), and I think changing it to "image-name" rather than tag would be less confusing… because the tag will actually have the repository plugged on the front.

Let me know if #28 looks good, and perhaps we can go with that instead of this one?

sjauld:
I'll do a quick test now.

sjauld:
Looks good to be, I'll close this one

toolmantim:
@sjauld cool! Thanks again for this… it wouldn't have happened without putting together this PR

PR #26, opened 2016-12-15
Save container logs as artifacts and output linked container logs on failure

lox:
This adds a step at the end of the run process that will show the output of containers if the main command fails, or if any of those containers have non-zero exit codes. 

Container output is uploaded as artifacts for easier viewing as well as the last 500 lines tailed in the output.


toolmantim:
Nice! 

I haven't had a chance to use it yet, but how big are container logs? I wonder if they're better piped to a `.log` file and uploaded as a build artifact? Or is putting it inline heaps nicer/faster for debugging? Or you we could tail the last 500 lines, and put the rest in a build artifact. 

It'd be great to get this to the point that it could be `onfail` by default.

lox:
Yup, I think I'm going to make it always save container output as artifacts, and then `tail -n 500` any of those logs where the container has failed.

This ditches any configuration and makes for sane behaviour I think. Thoughts @toolmantim?

toolmantim:
Nice!

On third review, I think perhaps it might be too aggressive to always post the container logs as artifacts? It could be a lot of unnecessary artifact storage, and mess up people's own artifact counts. Perhaps only on failure?

hooks/commands/run.sh
```diff
@@ -83,3 +84,45 @@ if [[ -f "$COMPOSE_SERVICE_OVERRIDE_FILE" ]]; then
...
+}
+
+if [[ $exitcode -ne 0 ]] ; then
+  echo "^^^ +++"
+  echo "+++ :warning: Failed to run command, exited with $exitcode"
+fi
+
+tail_failed_containers() {
+  for container_name in $(list_containers); do
+    container_exit_code=$(docker inspect --format='{{.State.ExitCode}}' "$container_name")
```
toolmantim: This is neat!

lox:
Any objections to me merging this @toolmantim @sj26 ?

hooks/commands/run.sh
```diff
@@ -2,6 +2,7 @@
 
 COMPOSE_SERVICE_NAME="$BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN"
 COMPOSE_SERVICE_OVERRIDE_FILE="docker-compose.buildkite-$COMPOSE_SERVICE_NAME-override.yml"
+LOGS_SETTING="${BUILDKITE_PLUGIN_DOCKER_COMPOSE_LOGS:-}"
```
toolmantim: Does this do anything? I can't see where it's used…

lox: Nope, will remove.

lox: Done.

PR #27, opened 2017-01-09
Remove support for docker-compose 1.6 and below

toolmantim:
Support for 1.6 and below seems broken as `remove_volume_flag` is undefined. And it's time we remove support for old docker-compose.


lox:
Looks great! Shall I merge this?

toolmantim:
@lox I just did! :)

Issue #29, opened 2017-01-10
BUILDKITE_PROJECT_SLUG has invalid character for Docker Image Tag


jufemaiz:
https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/hooks/commands/build.sh#L6

> A tag name may contain lowercase and uppercase characters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.
Source: https://docs.docker.com/engine/reference/commandline/tag/

```ruby
BUILDKITE_PROJECT_SLUG = [BUILDKITE_ORGANIZATION_SLUG, BUILDKITE_PIPELINE_SLUG].join('/')
```

Need to amend to use something like:

```bash
BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_NAME="${BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_NAME:-${BUILDKITE_ORGANIZATION_SLUG}-${BUILDKITE_PIPELINE_SLUG}-${BUILDKITE_PLUGIN_DOCKER_COMPOSE_BUILD}-build-${BUILDKITE_BUILD_NUMBER}}"

```

jufemaiz:
👍

PR #30, opened 2017-01-10
BUGFIX BUILDKITE_PROJECT_SLUG

jufemaiz:
As per https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/29


toolmantim:
Ah, good catch. Thanks for the PR.

The original intention was not to have the org in there at all, I just misspelt the env variable name. Could you update this to just use `BUILDKITE_PIPELINE_SUG` and not have the org in there at all, which will match the original code?

jufemaiz:
@toolmantim bump. Fixed.

toolmantim:
Thanks @jufemaiz!

Issue #37, opened 2017-04-14
Support for configurable Compose file version


dalberto:
I'm attempting to use this plugin to run a docker-compose pipeline. The application was configured to use docker-compose prior to the experiment with Buildkite. In order to support environment variable interpolation *with* defaults in the compose file, the application uses the 2.1 format. Unfortunately, it looks like this causes an error similar to the following:

```
ERROR: Version mismatch: file ./docker-compose.yml specifies version 2.1 but extension file ./docker-compose.buildkite-web-override.yml uses version 2.0
```

One potential solution is to support an environment variable that specifies the docker-compose version (which itself can default to 2.0 😄 ).  I imagine this would make it easier for existing applications with significant compose configurations to get onboard.

Another solution is to add an environment variable by disables the override altogether. This solution might be trickier, since (judging from `build.sh`), there would be no guarantee that the image is correctly tagged for pushing.

lox:
Ugh, sorry about that! Very annoying that they don't support 2.0 + 2.1 given the subtle differences. 

I think it would make sense to detect the version of the file and match that in the override. We'd love a PR, otherwise I can look at this when I get a chance.

lox:
Fixed in #40! Thanks!

PR #38, opened 2017-04-14
Detect docker-compose version

dalberto:
This PR seeks to address https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/37.

This change allowed me to run a pipeline with docker-compose 2.1 files. 

There are a few areas of improvement:
- [ ] There's probably a cleaner way to get the docker-compose version in a single command.
- [ ] This change assumes there is a `docker-compose.yml` file in the repository. I'm not sure how I can dynamically get the first compose file from the plugin configuration.

I'm happy to address this and any other issue or concern but a bit of guidance would be appreciated. 😄

### DIFF
--- a/hooks/commands/build.sh
+++ b/hooks/commands/build.sh
@@ -15,12 +15,14 @@ else
   TAG="$BUILDKITE_PLUGIN_DOCKER_COMPOSE_IMAGE_NAME"
 fi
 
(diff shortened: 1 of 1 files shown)


lox:
Thanks very much for the contribution @dalberto! I'll dive into it and provide some feedback.

lox:
It turned out that figuring out the config file was a bit more complicated than I though, so I combined the above in #40.

PR #40, opened 2017-04-14
Detect compose file version

lox:
This adds support for compose-file v2.1 by detecting the version from the first docker-compose config provided and using that for the overlay. 

This addresses #37. Many thanks to @dalberto for the report and initial implementation.


dalberto:
A similar change is necessary in `run.sh`, see: https://github.com/dalberto/docker-compose-buildkite-plugin/commit/53bbdba66041ed4fbc2a9f58c8c298b99cefb0fc

PR #42, opened 2017-04-21
Add bats tests

lox:
Whilst the "tests in docker-compose.yml" approach was working ok, it wasn't scaling well and it was hard to capture nuance. In this PR I've refactored the config parsing into a central function and I've added bats tests to reflect this. These are run first, and then the previous "smoke tests".

Functionally, this doesn't change anything. In terms of refactoring, it does the following:

 * Moves shared functions to `/lib/shared.bash`
 * Adds a `docker_compose_config_files` function that figures out a list of config files
 * Refactor `docker_compose_config_file` and `run_docker_compose` to use the new `docker_compose_config_files`
 * Rather than having epic long ENV vars, uses new `plugin_read_config` function
 * Lowercase vars for local stuff, uppercase for stuff from outside the script

I think the approach makes for more bash, but more maintainable. Thoughts @toolmantim?


toolmantim:
I forgot to mention, this is a really nice change and will help heaps when working on making any changes to the plugin. I've wanted some BATS tests multiple times.

PR #48, opened 2017-04-25
Support multiple image builds

lox:
This adds support for multiple image builds. The use-case for this is where you have multiple services that need building in one compose file:

```yml
steps:
  - name: ":docker: Build"
    agents:
      queue: docker-builder
    plugins:
      docker-compose:
        build: 
          - app
          - tests
        image-repository: index.docker.io/org/repo

  - wait

  - name: ":docker: Test"
    command: test.sh
    plugins:
      docker-compose:
        run: tests
```

This adds an additional metadata key that is numerically indexed, rather than by service name, so the run step can iterate through them. 

This also adds test coverage for build and run commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+docker-compose.buildkite-1-override.yml 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
 FROM lucor/bats
 
+ENV LIBS_BATS_MOCK_VERSION "1.0.1"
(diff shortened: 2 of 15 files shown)


toolmantim:
👍🏼

toolmantim:
Having tests for this is great!

toolmantim:
To release new versions of this we simply tag the version and update the readme. Let's switch to full sem-ver numbers tho.

lox:
Shall we merge it and put out a release @toolmantim?

toolmantim:
Yep!

PR #53, opened 2017-05-02
Add support for push command

lox:
This allows for a step that pushes your images to an upstream repository.

```yml
steps:
  - name: ":docker: Push to final repository"
    plugins:
      docker-compose#v1.2.1:
        push: 
        - app:index.docker.io/org/repo/myapp
        - app:index.docker.io/org/repo/myapp:latest
```

Interestingly, this was the most complicated of the three commands. The scenario where there are prebuilt images and you are wanting to tag and push those images gets complicated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ A [Buildkite](https://buildkite.com/) plugin allowing you to create a build syst
 * Containers are built, run and linked on demand using Docker Compose
 * Containers are namespaced to each build job, and cleaned up after use
 * Supports pre-building of images, allowing for fast parallel builds across distributed agents
+* Supports pushing images to a repository
 
 ## Example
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -149,13 +165,19 @@ Either a single service or multiple services can be provided as an array.
 
 The name of the service the command should be run within. If the docker-compose command would usually be `docker-compose run app test.sh` then the value would be `app`.
 
+### `push`
+
+A list of services to push in the format `service:image:tag`. If an image has been pre-built with the build step, that image will be re-tagged, otherwise docker-compose's built in push operation will be used. 
+
+Be aware that there is a race condition on tagging prebuilt images and pushing them if multiple push steps run in parallel. It's advisable to use a concurrency group in this situation.
```
toolmantim: This sounds scary! Is there anything specific about this plugin that introduces a race condition? Or is that a general "don't push two things at once"?

lox: Lol, yes. Agreed. That was very developer of me :) 

I will document this better, it's just the result of the fact that you can't tag and push atomically in docker, so it's theoretically possible if you had two parallel builds tagging `my.sweet.image:blah` and pushing it at the same time that you'd be pushing the wrong image. It's possible too academic.

PR #58, opened 2017-07-01
Add retry support to pull and push

lox:
This allows a `pull-retry: <int>` and `pull-retry: <int>`.

~~My concerns are that it's not immediately obvious what the retry is for. Thoughts @toolmantim?~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,14 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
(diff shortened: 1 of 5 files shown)


toolmantim:
What about?

```yml
push-retries: 5
pull-retries: 5
```

And then peeps can send it via an env if they want it to apply to everything all the time?

lox:
This should be good to go @toolmantim

PR #62, opened 2017-07-21
Support parallel builds

lox:
This unwinds some complexity that was added in #48. 

Where previously the only metadata that was set was `docker-compose-plugin-built-image-tag-${service}`, that PR added numerically indexed keys like `docker-compose-plugin-built-image-tag-0` and `docker-compose-plugin-built-image-count`.

The idea behind that was to allow subsequent run and pull steps to pull down ALL prebuilt images, rather than those that matched the name of the service in the run and pull step. 

The huge downside of the change introduced in #48 is that a) it creates a race condition if you have parallel build steps and b) generally you don't want to pull down all the images. 

This PR rolls us back to the `docker-compose-plugin-built-image-tag-${service}` keys, and drops support for "pull down all the things". In future, we can opt-in to pulling down extra pre-builts with something like a `pull` directive.

This should make parallel builds work again!


KevinGrandon:
> a) it creates a race condition if you have parallel build steps

Could you describe this a bit further? Anything we should be worried about?

lox:
Any build steps that are in parallel will race to set `docker-compose-plugin-built-image-count`. Until this is merged, I'd encourage running your builds in serial.

lox:
We'll get a release out today @KevinGrandon.

lox:
@toolmantim pointed out that my description was confusing @KevinGrandon. This PR removes the race condition, as opposed to introducing one 😅

KevinGrandon:
Got it, thanks! 👍

toolmantim:
That diff 😍

that-diff

PR #66, opened 2017-08-22
Build a service before run if there is no prebuilt image

lox:
Currently, without a `build` step, a `run` step doesn't force a build, which can lead to confusing stale image re-use. 

This allows the use of just a `run` step that works as you would expect.

### DIFF
--- a/hooks/commands/run.sh
+++ b/hooks/commands/run.sh
@@ -43,6 +43,7 @@ set +e
 if [[ -f "$override_file" ]]; then
   run_docker_compose -f "$override_file" run "$service_name" $BUILDKITE_COMMAND
 else
+  run_docker_compose build "$service_name"
   run_docker_compose run "$service_name" $BUILDKITE_COMMAND
 fi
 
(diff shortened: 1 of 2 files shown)


lox:
Took me a while to remember why @toolmantim, and I might still be mixed up. Imagine a docker-compose file like: 

```yaml
  helloworldimage:
    image: myhelloworld
    build: .
```

If you run this without building it then it will look for an image called `myhelloworld` and carry on if it exists, regardless of whether it's in sync with whatever is in `.`. We use that trick to pass the pre-built images in actually. 

The projectname would ensure that it worked in absence of the image, but when both are present and there is no pre-built you get whatever the latest `myhelloworld` image is on that host.

lox:
Actually, even if there isn't a `build` line, we are probably susceptible to stale images for `run` without prebuild too.

lox:
Yeah, I agree @toolmantim. Sorry it took me a while to articulate!

toolmantim:
No problemo! Good to have a record of the tradeoffs for future investigations.

PR #69, opened 2017-08-23
Update README for environment variable usage for pull and push retries.

KevinGrandon:
Currently it's not very clear that this is supported unless you do some digging. I think exposing this in the README could help folks.


toolmantim:
Thanks for the PR @KevinGrandon! I think eventually we'll let people know you can do this from some general plugins documentation, because it applies to any plugin configuration variable. But because these are probably going to be set a system level, it totally makes sense. Thanks again.

PR #72, opened 2017-09-05
Support env parameter to docker-compose

lox:
This adds support for an `env` block that is passed in to `docker-compse run`, for instance:

```yaml
steps:
  - command: test.sh
    plugins:
      docker-compose#v1.5.2:
        run: app
        env:
          - MYENV=1
          - MYENV=2
        config:
          - docker-compose.yml
          - docker-compose.test.yml
```

### DIFF
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -31,10 +31,14 @@ steps:
 
   - wait
   - label: run, with environment
+    command: "sh -c \"[ $$LLAMAS = always ] && [ $$TESTING = alpacas ]\""
     plugins:
       ${BUILDKITE_REPO}#${commit}:
         run: alpinewithenv
         config: tests/composefiles/docker-compose.v2.1.yml
+        environment:
+          - TESTING=alpacas
+
 
   - wait
   - label: build
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -30,6 +30,13 @@ steps:
         config: tests/composefiles/docker-compose.v2.1.yml
 
   - wait
+  - label: run, with environment
+    plugins:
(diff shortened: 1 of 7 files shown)


keithpitt:
I'll just get someone to test it before merge

toolmantim:
This is great! Do you know if this works with environment variable pass-through? I couldn't tell from the tests/docs.

lox:
Nope. I've been considering whether we should look at passing through some of the top level blocks like `artifact_paths` and `env`, but at present it doesn't, as there is no good way to do that and even if there was I'm not sure it's obvious. What are your feels?

toolmantim:
I only ask because I think you can do this via `docker-compose cli`:

```
$ env MOO=cow docker-compose run --entrypoint bash -e MOO agent
root@3edefa148d14:/go/src/github.com/buildkite/agent# echo $MOO
cow
```

which means you can probably do this via this plugin too?

lox:
OH! I totally misunderstood your question. Yes! You can do exactly that:

lox:
```
steps:
  - command: test.sh
    plugins:
      docker-compose#v1.5.2:
        run: app
        env:
          - LLAMAS
          - ALPACAS
        config:
          - docker-compose.yml
          - docker-compose.test.yml
```

lox:
That will pass through the value of LLAMAS or ALPACAS.

toolmantim:
Ok, cool! That's awesome. Sorry I should have been clearer to begin with.

Should we clarify that in the docs and tests? Because I reckon that's going to be a common use-case, especially for elastic-ci-stack peeps. Interestingly, the `docker-compose run --help` docs don't really get it right either!

lox:
Agreed @toolmantim, added some docs, let me know what you think.

PR #82, opened 2017-11-15
basic cache_from support

dreyks:
### Issue
To save resources we spin down all our runners and builders when not needed. However this busts docker caches for builders and that is bad. Starting from compose file `v3.2` there's a `cache_from` option that allows docker to use caches from any image present locally (image has to be pulled before it can be used as a cache source)

### Usage
```yml
plugins:
  docker-compose:
    build: app
    cache_from:
      - app:your.repo/image:tag
```

### Limitations

This is still kinda work in progress, but that'll do right now since `Bash` is not my native language and I've had enough of it for one day. I'll keep adding more commits to this PR

#### Todo
- [x] fail when trying to use `cache_from` with version < 3.2
- [x] refactor ugly duplicate code that reads cache_from
- [x] ~figure out how to use both remote and local caches simultaneously~ do not set cache_from if the image failed to pull
- [ ] (maybe not in this PR) support using several cache_from images for a single service

I'd be grateful if someone would read through this and point out my mistakes while I'm making them ;)

### DIFF
--- a/lib/shared.bash
+++ b/lib/shared.bash
@@ -100,9 +100,17 @@ function build_image_override_file_with_version() {
     IFS=':' read -a tokens <<< "$line"
     cache_from[${tokens[0]}]=$(IFS=':'; echo "${tokens[*]:1}")
   done
+  local using_cache_from=${cache_from[@]+"${#cache_from[@]}"}
 
   if [[ -z "$version" ]]; then
     echo "The 'build' option can only be used with Compose file versions 2.0 and above."
+    echo "For more information on Docker Compose configuration file versions, see:"
+    echo "https://docs.docker.com/compose/compose-file/compose-versioning/#versioning"
+    exit 1
+  fi
+
+  if [[ -n "$using_cache_from" && $(bc <<< "$version < 3.2") -gt 0 ]] ; then
+    echo "The 'cache_from' option can only be used with Compose file versions 3.2 and above."
     echo "For more information on Docker Compose configuration file versions, see:"
     echo "https://docs.docker.com/compose/compose-file/compose-versioning/#versioning"
     exit 1
--- a/lib/shared.bash
+++ b/lib/shared.bash
@@ -94,6 +94,12 @@ function build_image_override_file() {
 # docker-compose version and set of service and image pairs
 function build_image_override_file_with_version() {
   local version="$1"
+  declare -A cache_from
+
+  for line in $(plugin_read_list CACHE_FROM) ; do
+    IFS=':' read -a tokens <<< "$line"
+    cache_from[${tokens[0]}]=$(IFS=':'; echo "${tokens[*]:1}")
+  done
 
   if [[ -z "$version" ]]; then
     echo "The 'build' option can only be used with Compose file versions 2.0 and above."
@@ -102,6 +108,8 @@ function build_image_override_file_with_version() {
     exit 1
   fi
 
+  # TODO: cache_from requires version: 3.2
+
   printf "version: '%s'\n" "$version"
   printf "services:\n"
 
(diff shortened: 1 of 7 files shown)


lox:
This looks great! Thanks so much! I'll review in detail today.

Btw, just so you know, there are some minor gotcha's with `cache_from`: https://github.com/moby/moby/issues/32612 and it's also a relatively recent addition to docker-compose, but we can handle that with a version check I think.

dreyks:
Yes, that was a thing on my to-do list just forgot to mention it

dreyks:
so, I think I've finished. I probably don't want to implement the ability to have multiple caches for a single service right now

lib/shared.bash
```diff
@@ -102,14 +102,30 @@ function build_image_override_file_with_version() {
     exit 1
   fi
 
+  cache_from_not_available=($(bc <<< "$version < 3.2"))
```
lox: Unfortunately, I suspect we will have to do specific docker-compose version checks, rather than just the file version.

lox:
I'm a bit concerned about the lack of solid support in docker-compose for `cache_from`. It's hard to tell, but it looks like it's still pretty broken: https://github.com/docker/compose/issues/5090 https://github.com/docker/docker-py/issues/998 https://github.com/docker/docker-py/pull/1582

Are you seeing this working for you @dreyks?

dreyks:
hmm. it's working for me yes. a new agent is started, the latest image is pulled and then build uses all the caches

lib/shared.bash
dreyks: as per https://docs.docker.com/compose/compose-file/ I think this is the compose file version
```
CACHE_FROM

> Note: This option is new in v3.2
```

lox:
Well, if it works, I'm happy :)

lib/shared.bash
lox: Rather than bring in `bc`, you can use bash arithmatic contexts:

```
if (( a > b )); then
    ...
fi
```

lib/shared.bash
```diff
@@ -102,14 +102,30 @@ function build_image_override_file_with_version() {
...
   printf "services:\n"
 
   shift
   while test ${#} -gt 0 ; do
     printf "  %s:\n" "$1"
     printf "    image: %s\n" "$2"
-    shift 2
+
+    if [[ -n "$3" ]] ; then
+      if [[ $cache_from_not_available -gt 0 ]] ; then
```
lox: I generally use a "not null" check: `if [[ -n $cache_from_not_available ]]` for bool-like flags.

lib/shared.bash
lox: Unfortunately it was broken until docker-compose 1.14.0 for multiple files, but at least it didn't fail with an error. This check is probably ok.

dreyks: as far as I know bash arithmetics does not work with floats. I would have to either multiply it by something to make it integer or split it by the dot and compare parts manually. I'm not that good at bash though so maybe there's another way

lib/shared.bash
dreyks: :+1: will change

lox:
I'm good to merge this in, anything outstanding @dreyks?

dreyks:
I think this is good for the first implementation. I'll see if I can improve something when it's fully running on all our builds

lox:
Thanks for your hard work @dreyks!

toolmantim:
This great, thanks @dreyks. Any chance we could get.a readme update to document the new configuration option?

dreyks:
#83

toolmantim:
Looks like we broke a few platforms we previously supported (#88 #89). Will look into some fixes!

PR #87, opened 2018-01-15
Add a no-cache build option

toolmantim:
This adds a `no-cache: true` option to the build stage. It also cleans up some of the readme, and documents the `leave-volumes` option.


toolmantim:
This was released in v1.8.2

PR #93, opened 2018-01-18
do not use declare -A

dreyks:
do not use `declare -A` since it's not available in Bash 3

### DIFF
--- a/hooks/commands/build.sh
+++ b/hooks/commands/build.sh
@@ -5,7 +5,6 @@ image_repository="$(plugin_read_config IMAGE_REPOSITORY)"
 pull_retries="$(plugin_read_config PULL_RETRIES "0")"
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 build_images=()
-declare -A cache_from
 
 for line in $(plugin_read_list CACHE_FROM) ; do
   IFS=':' read -a tokens <<< "$line"
@@ -14,7 +13,7 @@ for line in $(plugin_read_list CACHE_FROM) ; do
 
   echo "~~~ :docker: Pulling cache image for $service_name"
   if retry "$pull_retries" plugin_prompt_and_run docker pull "$service_image" ; then
-    cache_from[$service_name]=$service_image
+    printf -v "cache_from__${service_name}" %s "$service_image"
   else
     echo "!!! :docker: Pull failed. $service_image will not be used as a cache for $service_name"
   fi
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks @dreyks! I'll add @lox as a reviewer because his Bash voodoo is strong

hooks/commands/build.sh
```diff
@@ -14,7 +13,7 @@ for line in $(plugin_read_list CACHE_FROM) ; do
 
   echo "~~~ :docker: Pulling cache image for $service_name"
   if retry "$pull_retries" plugin_prompt_and_run docker pull "$service_image" ; then
-    cache_from[$service_name]=$service_image
+    printf -v "cache_from__${service_name}" %s "$service_image"
```
lox: I had to look this up:

```
printf [-v var] format [arguments]

The -v option causes the output to be assigned to the variable var rather than being printed to the standard output.
```

lox: TIL :)

lox:
I think this approach is decent. I was going to suggest moving the "Is this service cached? If so pull it before build" into the main build loop, but I assume we do the pulling before the building because we might have services in cache_from that aren't in build?

hooks/commands/build.sh
dreyks: this is my google-driven development at its best :D

dreyks:
> we might have services in cache_from that aren't in build

um... I'm not sure I entirely understand this. If a service is not in build there's no need to pull it, cause it won't be used

lox:
Yeah, fair point. I was overthinking it! In that case, I'd suggest simplifying things down to one loop that builds images and pulls them if needed:

```bash
# process each service in the build directive
for service_name in $(plugin_read_list BUILD) ; do
  image_name=$(build_image_name "${service_name}")

  if [[ -n "$image_repository" ]]; then
    image_name="${image_repository}:${image_name}"
  fi

  # if there is a cache_from entry for the service we need to pull it down first
  if cache_from_image=$(cache_from_for_service "${service_name}") ; then

    echo "~~~ :docker: Pulling cache image for $service_name"
    if ! retry "$pull_retries" plugin_prompt_and_run docker pull "$cache_from_image" ; then
      echo "+++ :docker: Failed to pull image $cache_from_image"
      exit 1
    fi
    
    # the third parameter to a build_image tuple is the cache_from image to use, this is read by build_docker_blah_blah
    build_images+=("$service_name" "$image_name" "${cache_from_image}")
  else 
    build_images+=("$service_name" "$image_name" "")
  fi
done
```

lox:
That avoids the problem entirely of having to track which images have a cache_from entry.

lox:
I'm going to merge this as-is, because I need it for development on macOS :)

lox:
No problems, thanks again for your time @dreyks !

PR #95, opened 2018-01-23
Fix command quoting bug

### LINKED ISSUE #75: Using commands with this plugin doesn't appear to work
Using this plugin and this step:

```
  - label: build
    commands:
      - python setup.py sdist
      - buildkite-agent artifact upload dist/*.tar.gz
    plugins:
      docker-compose#v1.5.2:
        run: app
        config: docker-compose.buildkite.yml
```

I expected either two calls to docker-compose or a single call using bash eg:

`docker-compose -f docker-compose.buildkite.yml run app bash -c "python setup.py sdist && buildkite-agent artifact upload dist/*.tar.gz"`

Instead the command reported by buildkite is:

`docker-compose -f docker-compose.buildkite.yml -p buildkited85f2a76afb84b34ba66557c6bd4dc25 run app python setup.py sdist buildkite-agent artifact upload dist/\*.tar.gz`

Where the commands are concatenated without any handling.

lox:
Closes #75.


tests/run.bats
```diff
@@ -33,6 +33,31 @@ load '../lib/run'
...
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN=myservice
+  export BUILDKITE_PIPELINE_SLUG=test
+  export BUILDKITE_BUILD_NUMBER=1
+  export BUILDKITE_COMMAND="sh -c 'echo hello world'"
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CHECK_LINKED_CONTAINERS=false
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLEANUP=false
+
+  stub docker-compose \
+    "-f docker-compose.yml -p buildkite1111 build --pull myservice : echo built myservice" \
+    "-f docker-compose.yml -p buildkite1111 run --name buildkite1111_myservice_build_1 myservice $BUILDKITE_COMMAND : echo ran myservice"
```
toolmantim: Maybe it's worth copy+pasting the actual command into the expectations, rather than using $ substitution, only because the test is related to $ substitution?

lox: Yeah 🤔 I kept running into copy-paste errors was why I did that! 😅

toolmantim: 😂

PR #103, opened 2018-03-06
Correctly capture failures from the Subshell

Sutto:
Currently the -e option is turned off from inside the subshell - This fails though, as it remains on for the parent shell.

Thus, to capture (and handle) failures, we need to turn the option off in the top level shell correctly, rather than in a child shell, and restore this *after* our block of code has been run.

This manifests itself in the form of failed commands only invoking cleanup, and none of the rest of the code. The check on exit code is completely skipped and, for instance, docker logs from linked containers are never uploaded (which is far from ideal).

### DIFF
--- a/hooks/commands/run.sh
+++ b/hooks/commands/run.sh
@@ -57,10 +57,13 @@ fi
 
 run_params+=("$service_name")
 
+# Disable -e outside of the subshell; since the subshell returning a failure
+# would exit the parent shell (here) early.
+set +e
+
 (
   # Reset bash to the default IFS with no glob expanding and no failing on error
   unset IFS
-  set +e
   set -f
 
   # The eval statements below are used to allow $BUILDKITE_COMMAND to be interpolated correctly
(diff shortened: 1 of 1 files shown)


lox:
Thanks for this! I noticed that this was happening the other day and was trying to write some tests to prove what was going on. I'm going to try and commit those tests first to prove the breakage.

lox:
I'm going to merge this and add tests later. Thanks again @Sutto.

Sutto:
No problem! Thanks for looking at it so quickly. I also noticed that it's possible for the docker logs to fail in some cases - so it might be worth wrapping that with the same trick / ignoring their exit code (or else cleanup will be invoked early).

lox:
> No problem! Thanks for looking at it so quickly. I also noticed that it's possible for the docker logs to fail in some cases - so it might be worth wrapping that with the same trick / ignoring their exit code (or else cleanup will be invoked early).

Do you mean `check_linked_containers_and_save_logs`?

Sutto:
Sorry, I missed your reply - yep. It's possible if a container had no output that it will exit with a non-zero error code - and likewise, it double prints the output (since it runs and prints to stdout, then uploads the artifact) - for the main container run.

lox:
Ah yup, I noticed that too. I'll look into it. Cheers @Sutto.

PR #106, opened 2018-03-07
Support pull_all directive for pulling all images

### LINKED ISSUE #86: Ability to pull multiple images in run step
Hi,

I'm attempting to run tests of our whole suite (our clients and servers are in separate repos).

To do this, I've created a parent repo with the other repos as submodules. I've created a pipeline and docker-compose files in the parent repo so that Buildkite can build and run the child project images.

I would like to pre-build the images but find that when it hits the run step, it only pulls the image that I'm running, and then proceeds to rebuild the other images specified in docker-compose.yml.

I had a quick look through the plugin code and found this comment:

> We only look for a prebuilt image for the service being run. This means that
> any other services that are dependencies that need to be built will be built
> on-demand in this step, even if they were prebuilt in an earlier step.

Is this something that's likely to change?

Thanks
Ritchie

lox:
Via https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/86#issuecomment-368724099

Closes #86.

### DIFF
--- a/hooks/commands/run.sh
+++ b/hooks/commands/run.sh
@@ -19,6 +19,12 @@ if [[ "$(plugin_read_config CLEANUP "true")" == "true" ]] ; then
   trap cleanup EXIT
 fi
 
+# We support pulling all images, in case they have been pulled on the agent machine already in
+# an earlier build, and need to be updated
+if [[ "$(plugin_read_config PULL_ALL "false")" == "true" ]] ; then
+    retry "$pull_retries" run_docker_compose pull --parallel
+fi
+
 test -f "$override_file" && rm "$override_file"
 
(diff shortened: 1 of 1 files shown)


lox:
I like the simplicity of this approach @dominics.

lox:
Hmmm, unfortunately this won't work. This is doing a `pull` before the override file is created, which means it will pull whatever images are referenced in the the original `docker-compose.yml`, not the prebuilt images. Have I misunderstood your intention here @dominics?

dominics:
Nah, no misunderstanding. In my particular (peculiar?) use case, I didn't need prebuilts when pulling all. (It was for an e2e test where I just wanted all the original docker-compose services pulled up to date)

PR #110, opened 2018-03-07
Support artifact_paths inside containers

lox:
This mounts any artifact paths you specify into the container so that files written there will make it back to the host for upload.

### DIFF
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -40,6 +40,14 @@ steps:
         environment:
           - ALPACAS=sometimes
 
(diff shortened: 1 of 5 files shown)


toolmantim:
Neat idea! I think the code still has `dist/*` hard coded.

lox:
Yeah, was just resurrecting a very old branch @toolmantim, I still don't have a good way of making this work. I think I'm going to take a different approach and try and volume mount in the artifact_paths

lox:
I never figured out a good way to do this.

PR #112, opened 2018-03-15
Scope build meta-data keys by config file

### LINKED ISSUE #111: Clashes when pre-building different docker-compose configs with the same service name
If you have the same service name in two different docker-compose config files, and try to pre-build them for later steps, only the last completed step will be used.

For example, given the following pipeline both of the test commands will run against the prod image, instead of running against their respective images:

```yml
steps:
  - name: ":docker:"
    plugins:
      docker-compose#v1.8.4:
        build: app
        image-repository: org/app-dev

  - label: ":docker::zap:"
    plugins:
      docker-compose#v1.8.4:
        config: docker-compose.prod.yml
        build: app
        image-repository: org/app-prod

  - wait

  - label: ":hammer:"
    command: make test
    plugins:
      docker-compose#v1.8.4:
        run: app

  - label: ":hammer::zap:"
    command: make test
    plugins:
      docker-compose#v1.8.4:
        config: docker-compose.prod.yml
        run: app
```

Perhaps the meta-data key we use to the store the image name should be scoped to the the docker-compose config file as well, and not just the service name?

toolmantim:
This updates the meta-data key used for storing pre-built images to include the docker-compose config file name. This allows you to pre-build and run the same named service from two completely different docker-compose.yml files, without any fear of clashing.

For example:

```yml
steps:
  - name: ":docker: dev"
    plugins:
      docker-compose#x:
        build: app
        image-repository: org/app-dev

  - label: ":docker: staging"
    plugins:
      docker-compose#x:
        config: docker-compose.staging.yml
        build: app
        image-repository: org/app-staging

  - wait

  - name: ":hammer: dev"
    command: yarn test
    plugins:
      docker-compose#x:
        run: app

  - name: ":hammer: staging"
    command: yarn test-staging
    plugins:
      docker-compose#x:
        config: docker-compose.staging.yml
        run: app
```

The above two will now set and get the following meta-data keys:

```
docker-compose-plugin-built-image-tag-app
docker-compose-plugin-built-image-tag-app-docker-compose.staging.yml
```

If you specify multiple configs with the array syntax they're joined with a dash. For example, the following:

```yml
steps:
  - name: ":docker: dev"
    plugins:
      docker-compose#x:
        config:
          - docker-compose.one.yml
          - docker-compose.two.yml
        build: app
        image-repository: org/app-dev
```

Sets the following meta-data key:

```
docker-compose-plugin-built-image-tag-app-docker-compose.one.yml-docker-compose.two.yml
```

I don't think this should have any backwards compatibility problems, unless people were depending on the meta-data key, but as that's not configurable I'd consider it private.

- [x] Tests
- [x] Code

Fixes #111

### DIFF
--- a/hooks/commands/build.sh
+++ b/hooks/commands/build.sh
@@ -62,7 +62,7 @@ if [[ -n "$image_repository" ]]; then
   run_docker_compose -f "$override_file" push "${services[@]}"
 
   while [[ ${#build_images[@]} -gt 0 ]] ; do
-    plugin_set_metadata "built-image-tag-${build_images[0]}" "${build_images[1]}"
+    set_prebuilt_image "${build_images[0]}" "${build_images[1]}"
     build_images=("${build_images[@]:3}")
   done
 fi
--- a/lib/metadata.bash
+++ b/lib/metadata.bash
@@ -19,8 +19,20 @@ function plugin_set_metadata() {
   plugin_prompt_and_must_run buildkite-agent meta-data set "$key" "$value"
 }
 
-# Gets a prebuilt iamge for a service name
+function prebuilt_image_meta_data_key() {
+  local service="$1"
+  echo "${META_IMAGE_TAG}${service}"
+}
+
+# Sets a prebuilt image for a service name
+function set_prebuilt_image() {
+  local service="$1"
+  local image="$2"
+  plugin_set_metadata "$(prebuilt_image_meta_data_key "$service")" "$image"
+}
+
+# Gets a prebuilt image for a service name
 function get_prebuilt_image() {
   local service="$1"
-  plugin_get_metadata "${META_IMAGE_TAG}${service}"
+  plugin_get_metadata "$(prebuilt_image_meta_data_key "$service")"
 }
--- a/lib/metadata.bash
+++ b/lib/metadata.bash
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-readonly META_IMAGE_TAG=built-image-tag-
-
 # Read agent metadata for the plugin
 function plugin_get_metadata() {
   local key="docker-compose-plugin-$1"
@@ -19,9 +17,23 @@ function plugin_set_metadata() {
   plugin_prompt_and_must_run buildkite-agent meta-data set "$key" "$value"
 }
 
+# The service name, and the docker-compose config files, are the uniqueness key
+# for the pre-built image meta-data tag
 function prebuilt_image_meta_data_key() {
   local service="$1"
-  echo "${META_IMAGE_TAG}${service}"
+  local config_key=""
+
+  for file in $(docker_compose_config_files) ; do
+    config_key+="-$file"
+  done
+
+  # If they just use the default config, we use the old-style (non-suffixed)
+  # style key
+  if [[ "$config_key" == "-docker-compose.yml" ]]; then
+    echo "built-image-tag-$service"
+  else
+    echo "built-image-tag-$service$config_key"
+  fi
 }
 
 # Sets a prebuilt image for a service name
--- a/tests/build.bats
+++ b/tests/build.bats
@@ -64,6 +64,57 @@ load '../lib/shared'
   unstub buildkite-agent
 }
 
+@test "Build with a repository and custom config file" {
+  export BUILDKITE_JOB_ID=1111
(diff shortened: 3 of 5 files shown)


lox:
Gnarly problem!

lox:
> I don't think this should have any backwards compatibility problems, unless people were depending on the meta-data key, but as that's not configurable I'd consider it private.

Unfortunately, whilst this is compatible with the same version, if you mix and match accidentally (which is quite easy to do), it's API incompatible with previous versions.  

```
Running plugin github.com/buildkite-plugins/docker-compose-buildkite-plugin command hook | 0s
-- | --
  | # Executing "/tmp/buildkite-agent-bootstrap-hook-runner-856866600"
  | $ /bin/bash -c /tmp/buildkite-agent-bootstrap-hook-runner-856866600
  | $ buildkite-agent meta-data get docker-compose-plugin-built-image-tag-everchron-api-.buildkite/compose.yml
  | 2018-03-30 09:56:44 FATAL  Failed to get meta-data: POST https://agent.buildkite.com/v3/jobs/xxx/data/get: 404 No key "docker-compose-plugin-built-image-tag-ev..." found
  |   | Failed to get metdata docker-compose-plugin-built-image-tag-myapp-api-.buildkite/compose.yml (exit 1)
```

IMO we need to make this at least work ok with the 1.8.x series of metadata keys, and then probably bump the version to v2.0.0 to release this.

toolmantim:
Oh, interesting! I hadn’t really considered that.

Yeah, we should bump to 2.0

Also, we should change the output message as well… that meta-data key message is obtuse, and could be much friendlier.

toolmantim:
We haven’t cut a new version yet, I was hestitant to. Glad I didn’t!

Did this crop up for someone who was accidentally using `master`? (Which is perhaps another problem. 99% of the time you want it pinned, and all steps pinned to the same version).

/me ponders a GitHub bot that sends PRs to lint and update your BK pipeline plugin version deps.

toolmantim:
Any reason we need the backward compat? Why not cut v2.0.0 and in the release notes mention it?

jufemaiz:
Uhoh. I think v2.0.0 broke our builds :(

lox:
More context required @jufemaiz ;)

jufemaiz:
@lox confirmation on https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/112#issuecomment-377728994

Release notes didn't comment on breaking change and that's the exact message (within reason) I saw by bumping to v2.0.0. v1.8.4 works fine.

We are having to do intermediary build injecting agent held SSH creds via build arguments in so that we can access private git repo so we don't do a docker-compose plugin based build but obviously need to tag the metadata for future uses by docker-compose plugin.

```
buildkite-agent meta-data set docker-compose-plugin-built-image-tag-app $AWS_ACCOUNT.dkr.ecr.$MASTER_AWS_REGION.amazonaws.com/$REPOSITORY:$DEPLOY_VERSION
```

Using v2.0.0 plugin threw the following:

```
# Executing "/tmp/buildkite-agent-bootstrap-hook-runner-617244467"
--
  | $ /bin/bash -c /tmp/buildkite-agent-bootstrap-hook-runner-617244467
  | $ buildkite-agent meta-data get docker-compose-plugin-built-image-tag-app-docker-compose.test.yml
  | 2018-04-19 00:11:02 FATAL  Failed to get meta-data: POST https://agent.buildkite.com/v3/jobs/c9b435e1-94fe-446f-bb7d-96c1850c2ef7/data/get: 404 No key "docker-compose-plugin-built-image-tag-ap..." found
  |  
```
Which then attempted to do a standard docker-compose build (but no build arguments so no dice).

Would be good to know what the key is there to attempt to further debug.

lox:
Sorry about that @jufemaiz! We're trying to avoid having the metadata key schemes being a public API, but we should have been clearer about that. 

Might multi-stage docker builds be a good solution for you? https://medium.com/@kevinsimper/how-we-do-docker-multi-stage-builds-and-secure-sharing-of-private-repositories-956eda84f3b8

Alternately, we could prioritize build args to the docker-compose plugin.

jufemaiz:
No worries @lox. Would you like me to open as a separate issue?

lox:
That would be great if you could @jufemaiz!

PR #117, opened 2018-04-03
Allow specifying the --no-deps flag

tessereth:
I'm currently trying to migrate my buildkite pipeline to use the new plugin system but I've run into a little snag. I have some steps that require a bunch of dependent services but some steps that do not. I tried using the docker plugin for the steps that don't need the services but the docker plugin makes some assumptions that don't work in my case^. Given I have a docker compose file anyway and I'm using it for the other steps, it would be convenient for me to use the docker-compose plugin for everything but specify the `--no-deps` flag somehow. I'm not fussed _how_ that option gets set but given I made this change to confirm this solved my problem, I figured I'd make a PR rather than just an issue.

^ For completeness, the docker plugin always specifies the `--workdir` commandline flag, overriding the default I set in my `Dockerfile` and mounts the pwd at that directory which I also don't want. My source code is already inside the image that I prebuilt in a previous step. Theoretically changes could be made on the docker plugin to solve my problem as well but this seems cleaner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,12 @@ If set to false, doesn't allocate a TTY. This is useful in some situations where
 
 The default is `true`.
 
+### `dependencies` (optional, run only)
+
+If set to false, doesn't start linked services.
+
+The default is `true`.
+
 ### `verbose` (optional)
 
 Sets `docker-compose` to run with `--verbose`
--- a/hooks/commands/run.sh
+++ b/hooks/commands/run.sh
@@ -91,6 +91,11 @@ if [[ "$(plugin_read_config TTY "true")" == "false" ]] ; then
   run_params+=(-T)
 fi
 
+# Optionally disable dependencies
+if [[ "$(plugin_read_config DEPENDENCIES "true")" == "false" ]] ; then
+  run_params+=(--no-deps)
+fi
+
 run_params+=("$run_service")
 
 if [[ ! -f "$override_file" ]]; then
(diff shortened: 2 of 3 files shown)


lox:
This seems very reasonable. I'm pretty in favor of adding most of the `docker-compose run` args. Thoughts @toolmantim?

lox:
> ^ For completeness, the docker plugin always specifies the --workdir commandline flag, overriding the default I set in my Dockerfile and mounts the pwd at that directory which I also don't want. My source code is already inside the image that I prebuilt in a previous step. Theoretically changes could be made on the docker plugin to solve my problem as well but this seems cleaner.

We'll also have a think about this! Reckon that's a bug.

toolmantim:
@lox yeah, I reckon we should support all of the flags!

lox:
@toolmantim Should we implement them all as top level options, or add a compose_args option that passes them directly through?

toolmantim:
@lox top level options is nicer for the end-user, and the pipeline.yml, so I say that's worth the technical cost?

lox:
Agreed!

toolmantim:
(there's a only a couple more to add, as some don't make sense like `--detach`)

tessereth:
Sweet, thanks for merging this!

lox:
Thank you very much for your contribution @tessereth, we will get this into the next version 🙌🏻

PR #119, opened 2018-04-07
Add a JSON schema for the config options

toolmantim:
This adds a [JSON schema](http://json-schema.org) for the config options we accept for the plugin. This plugin is the most complex one we have so far, so I thought it'd be a good test case. Simpler plugins would have much simpler schemas.

All the examples in the readme validate, according to https://jsonschemalint.com/#/version/draft-06/markup/yaml

### DIFF
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -18,6 +18,15 @@ steps:
   - label: shellcheck
     command: .buildkite/steps/shellcheck
 
+  - label: lint
+    plugins:
+      docker#v1.1.1:
+        image: buildkite/plugin-linter
+        workdir: /plugin
+        environment:
+          - PLUGIN_NAME=docker-compose
+        always-pull: true
+
   - label: run bats tests
     plugins:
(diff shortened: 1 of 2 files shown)


toolmantim:
Once you have the schema, you can do neat things, like validate the usage examples in the readme:
https://gist.github.com/toolmantim/6ec7b42d4a388965645418e17b600edc

toolmantim:
I added a readme example validator (using https://github.com/buildkite/buildkite-plugin-example-validator) using the [ajv](https://www.npmjs.com/package/ajv) node package in a minimal Docker container. Ajv seemed to have the best error messages and schema support out of the Ruby/Node bunch.

Without a readme markdown config generator, or other way to view the config option docs, it wasn’t worth duping all the descriptions/defaults/examples in the schema.

But in the future, it’d be great if people didn’t have to maintain that config section of the readme.

I think this is good to merge?

toolmantim:
We now have https://github.com/buildkite-plugins/buildkite-plugin-linter, and updated this to use that image instead! I'll merge this now.

PR #121, opened 2018-04-09
Use the linter plugin

toolmantim:
Now there's a [Plugin Linter plugin](https://github.com/buildkite-plugins/plugin-linter-buildkite-plugin) 😱


lox:
🤯

PR #124, opened 2018-04-16
Run all commands through a shell interpreter

lox:
Multiple commands get passed through as new-line delimited commands. Docker needs a single command, so this currently doesn't work. This sets a standard command to execute $BUILDKITE_COMMAND in of `/bin/sh -e -c`, you can provide an alternative via the `shell` parameter. 

The downside is that this are:
* Adds a layer of shell execution to your docker-compose runs
* Requires `/bin/sh` in your container

We could possibly only do this if there is multiple commands, but in some ways that just makes it more magical IMO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,12 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
+### `shell` (optional, run only)
+
+The shell that is used to invoke commands. This is so that multiple commands can be interpretted. if `false` is specified, the command is invoked directly.
+
+The default is `/bin/sh -e -c`.
+
 ## Developing
 
 To run the tests:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -49,6 +49,7 @@ steps:
     plugins:
       ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         run: helloworld
+        shell: false
         config: tests/composefiles/docker-compose.v2.0.yml
 
   - wait
@@ -57,6 +58,7 @@ steps:
     plugins:
       ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         run: helloworld
+        shell: false
         config: tests/composefiles/docker-compose.v2.1.yml
 
   - wait
@@ -64,6 +66,7 @@ steps:
     plugins:
       ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         run: helloworld
+        shell: false
         config: tests/composefiles/docker-compose.v2.1.yml
 
   - wait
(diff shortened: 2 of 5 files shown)


hooks/commands/run.sh
```diff
@@ -8,6 +8,7 @@ run_service="$(plugin_read_config RUN)"
 container_name="$(docker_compose_project_name)_${run_service}_build_${BUILDKITE_BUILD_NUMBER}"
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 pull_retries="$(plugin_read_config PULL_RETRIES "0")"
+shell="$(plugin_read_config SHELL "/bin/bash -c")"
```
toolmantim: I think we have `-e` in the vanilla bootstrap command hook, to ensure that commands don't keep executing if one fails

toolmantim: Could we just default to `/bin/sh`, which on most systems will just be `bash` anyway?

lox: Yup, I reckon your are totes correct on both calls. As always you are a fine human to collaborate with.

.buildkite/pipeline.sh
```diff
@@ -42,9 +43,23 @@ steps:
...
         environment:
           - ALPACAS=sometimes
 
+  - wait
+  - label: run, with multiple commands
+    command:
+      - echo \$ALPACAS
+      - pwd
+      - echo "llamas rock too"
+    shell: "/bin/sh -c"
```
toolmantim: I think this needs to be under the plugin block?

lox:
Ugh, so how do we handle containers without a shell in them?

toolmantim:
Yeah, that's the same trouble I had trying to use the docker plugin with the [plugin-linter](https://github.com/buildkite-plugins/buildkite-plugin-linter) and trying to pass it command line args. Hrmmmm…

lox:
`shell: ~` ?

toolmantim:
`shell: false` ?

lox:
I guess you'd never want to use `/bin/false` as your shell 😅

README.md
```diff
@@ -293,7 +293,7 @@ The default is `false`.
 
 ### `shell` (optional, run only)
 
-The shell that is used to invoke commands. This is so that multiple commands can be interpretted.
+The shell that is used to invoke commands. This is so that multiple commands can be interpretted. if `false` is specified, the command is invoked directly.
```
toolmantim: The shell that is used to invoke commands. Set to `false` to disable running the commands via a shell (useful for images that expect command arguments to passed directly).

lox:
I'm going to close this for now, I think some more thinking is needed around how to handle multiple commands and containers without shells.

PR #125, opened 2018-04-16
Add docker-login example

KevinGrandon:
You can see the rendered markdown here: https://github.com/KevinGrandon/docker-compose-buildkite-plugin/blob/update-readme-auth/README.md#example


lox:
Thanks @KevinGrandon!

Issue #122, opened 2018-04-11
Unable to re-use pulled image specified by docker-compose file


KevinGrandon:
We have a base image which we use for several services in docker-compose.yml. This works great locally, but we haven't found a way that we can leverage this in Buildkite. We are able to pull the base image using `pull`, but it seems it can't find the named image that we pull down, and will build again unnecessarily.

Docker-compose file:

```yaml
version: '3.1'
services:
  base_service:
    build: .
    image: myimage

  service1:
    image: myimage:latest
    depends_on:
      - base_service

  service2:
    image: myimage:latest
    depends_on:
      - base_service
```

Pipeline looks like:

```yaml
steps:
  - name: ':docker: :package:'
    plugins:
      'docker-compose#v2.0.0':
        build: base_service
        image-repository: ...
  - wait
  - name: 'task 1'
    command: ./dosomething1
    plugins:
      'docker-compose#v2.0.0':
        run: service1
        pull:
          - base_service
  - name: 'task 2'
    command: ./dosomething2
    plugins:
      'docker-compose#v2.0.0':
        run: service2
        pull:
          - base_service
```

KevinGrandon:
Ok, so I have kind of a crazy workaround that works both locally and in CI for now, but it would be great if this worked out of the box. Posting here in case it helps and if anyone else needs a workaround. Here's what we're doing:

**New environment variable in the Buildkite UI:**
```
IMAGE_NAME_PREFIX=myrepo.amazonaws.com/myimage:myimage-base_service-build-
```

**Bash variable expansion in docker-compose that only works in CI:**
```
image: ${IMAGE_NAME_PREFIX:-myimage:latest}${BUILDKITE_BUILD_NUMBER:-}
```

This is a really nasty hack though, and if we could get this working with simply `image: myimage:latest` that would be ideal.

toolmantim:
Oh, interesting problem!

Ah, named images in docker-compose aren't scoped at all… so I can't think of a way where this would work, because you still need this to work:

```
  - name: ':docker: :package:'
    plugins:
      'docker-compose#v2.0.0':
        build: base_service
        image-repository: my.repo/image
```

This pushes to `my.repo/image:docker-compose-plugin-${pipeline-slug}-${build-number}`

And then the following run step:

```yaml
  - name: 'task 1'
    command: ./dosomething1
    plugins:
      'docker-compose#v2.0.0':
        run: service1
        pull:
          - base_service
```

[...]

PR #107, opened 2018-03-07
Support multiple values for image-names to match multiple build images

lox:
Closes #84.

PR #127, opened 2018-04-21
Add run-image-alias functionality

KevinGrandon:
The new configuration, run-image-alias will pull and alias an image for use by pipeline steps. This is one possible way that we could implement #122.

The goal of this new configuration is to generate the override file that looks like this:
```yaml
version: '3.1'
services:
  base_service:
    image: my.repo/image:docker-compose-plugin-${pipeline-slug}-${build-number}
  # Uses the built image for base_image, because we define run-image-alias in the pipeline.
  service1:
    image: my.repo/image:docker-compose-plugin-${pipeline-slug}-${build-number}
```

This probably isn't finished yet as I'm not great with bash, but opening this PR to get comments/thoughts. Maybe it's possible to leverage cache_from instead? Feel free to takeover/modify this PR.

### DIFF
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -119,18 +119,17 @@ steps:
         config: tests/composefiles/docker-compose.v2.1.yml
 
   - wait
-  - label: build with alias-from
+  - label: build with run-image-alias
     plugins:
       ${BUILDKITE_REPO}#${commit}:
         build: base_service
         config: tests/composefiles/docker-compose.v3.2.shared-base-service.yml
 
   - wait
-  - label: run with alias-from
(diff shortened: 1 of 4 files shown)


KevinGrandon:
@toolmantim @lox - I was able to get this working with a new config. Would appreciate any feedback on this if you have it. If something like this looks ok to you, I would appreciate any help getting this landed. Thanks!

lox:
It dawns on me that we originally added the `image` parameter in to handle something like this. I wonder if we merged #107 whether your example would work without the addition of a new concept.

lox:
Going to close this one for now, still figuring out how to solve this.

PR #129, opened 2018-04-23
Support cache-from in run

lox:
This allows for the `cache-from` directive to be used in run commands. This means you can prebuilt a shared base image and then use it as the cache ancestry for subsequent run commands that would otherwise rebuild images from scratch, for instance:

### pipeline.yml
```yaml
steps:
  - name: ':docker: :package:'
    plugins:
      'docker-compose#aa08b09':
        build: base_service
        image-repository: my.registry/llamas
  - wait
  - name: 'task 1'
    command: echo hello world
    plugins:
      'docker-compose#aa08b09':
        run: service1
        cache_from: base_service
  - name: 'task 2'
    command: echo hello world
    plugins:
      'docker-compose#aa08b09':
        run: service2
        cache_from: base_service
```

### docker-compose.yml
```
version: '3.4'
services:
  base_service:
    build: .

  service1:
    build: .

  service2:
    build: .
```

Related to #122 and #127.

### DIFF
--- a/hooks/commands/build.sh
+++ b/hooks/commands/build.sh
@@ -6,6 +6,7 @@ pull_retries="$(plugin_read_config PULL_RETRIES "0")"
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 build_images=()
 
+# Handle cache-from directive
 for line in $(plugin_read_list CACHE_FROM) ; do
   IFS=':' read -r -a tokens <<< "$line"
   service_name=${tokens[0]}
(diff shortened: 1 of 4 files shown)


lox:
Thoughts on this approach @KevinGrandon?

toolmantim:
This is also quite a handy standalone feature!

lox:
It is, at the cost of 80 lines of bash. It made me a bit sad.

KevinGrandon:
Looks like this could totally work, well done! I assume this will generate the service overrides properly?  Happy to give this a spin once released, or if there's a way that I can point my pipeline to a git ref I could try it earlier as well.

lox:
@KevinGrandon you should be able to try a specific plugin git ref via the syntax like in the example: `docker-compose#aa08b09`. One caveat to be aware of is that multi-stage builds don't work with cache-from. My initial attempt used them with the docker-compose `target` directive (which is super cool).

toolmantim:
How’d you go with this @KevinGrandon?

KevinGrandon:
Thank you for putting this together. I tested this and it seems to not work, I get an error like:

```
Service <service> has neither an image nor a build context specified. At least one must be provided.
```

It seems that this does not generate a valid override file.

toolmantim:
Dang, sorry about that @KevinGrandon. Will investigate.

toolmantim:
Removing my review on this, until we figure out what we want to do.

lox:
Closing for now, there hasn't been a strong need for this, and it adds a lot of complexity.

PR #132, opened 2018-05-02
Switch to a simpler pipeline.yml upload

toolmantim:
Pipeline upload now interpolates the plugin key too, so no need for these old tricks.


toolmantim:
@lox this makes for a nicer pattern for supporting self-testing `HEAD` commits… means that Renovate and tools like that can auto-update the pipeline.yml still…

toolmantim:
(all thanks to the changes in env var substitution improvements that were made for 3.0)

lox:
Yup, makes sense to me!

PR #136, opened 2018-05-03
Expands the previous header groups on error

lox:
Currently if a step like "Building" or "Pulling" fails, the error is buried because our header expansion gear favours the final cleanup step.

This ensures that on errors the previous group gets expanded before the cleanup step.


toolmantim:
I just gave this a go, and seems to work well!

PR #140, opened 2018-05-17
Fix bug with hyphens in service names

patrobinson:
We've found that if we have a docker-compose file like so:

```
services:
  app:
    build: ../..
    depends_on:
      - foo-db
  foo-db:
    image: postgres:9.4
```

The push command will tag the app image as `postgres:9.4`, which causes all kinds of bizarre problems.

The issue is the regular expression being used `\w` expands to `[_[:alnum:]]` which won't match a hyphen.

This PR fixes the issue and creates a test for the scenario

### DIFF
--- a/lib/push.bash
+++ b/lib/push.bash
@@ -5,7 +5,7 @@ compose_image_for_service() {
   local image=""
 
   image=$(run_docker_compose config \
-    | grep -E "^(  \\w+:|    image:)" \
+    | grep -E "^(  [_[:alnum:]-]+:|    image:)" \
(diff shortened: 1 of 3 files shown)


toolmantim:
Sorry you hit this bug. Thanks so much for the pull request… looks great!

toolmantim:
This has been released in [v2.2.1 (🐢 Hyphenation Turtle)](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/releases/tag/v2.2.1)

PR #142, opened 2018-05-18
Less alarming messages

sj26:
Calling out a header with "Failed" is leading some customers astray:

screen shot 2018-05-18 at 11 28 13 am

Expanding this header isn't useful:

screen shot 2018-05-18 at 11 28 25 am

The buildkite-agent meta-data commands are already quite chatty and should have sufficient output for debugging, without being alarmist about missing meta-data:

screen shot 2018-05-18 at 11 27 49 am

I've removed the header. Thoughts?


lox:
The message use to be "No prebuilt docker image found in meta-data", or something similar. IMO we should go back to that.

toolmantim:
Let's 🚀

PR #144, opened 2018-05-24
Fix 'environent' typo

clee:


toolmantim:
Thanks @clee - good catch!

PR #149, opened 2018-06-02
Add volume support for run

### LINKED ISSUE #148: volume support
It would be really great if we could specify volumes alongside env vars for each `run` service.

clee:
Fixes #148.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ volumes:
   - "./dist:/app/dist"
 ```
 
-If your Docker Compose config doesn't have the volume mount, and you can't add it for some reason, then you can specify a `volumes` directive in your pipeline using the same syntax for each mount; just make sure the `volumes` key is a sibling of the `run` key. Using the earlier example, this is how it would look to specify the volumes for the plugin:
+You can also use the `volumes` plugin option to add or override a volume, for example:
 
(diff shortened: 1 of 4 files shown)


lox:
Thanks for this @clee! I'm ok to add this, would it be possible to add a README entry? Ideally we'd get a test too, but I can help with that if need be.

clee:
Good call, @lox! This PR has been updated with new documentation in the README. I attempted to match the tone and style of the existing documentation, but I'm happy to adjust it (or for y'all to change it).

toolmantim:
Thanks for the PR @clee! I've just rebased this, added some tests for you.

One thing I noticed, is that it seems to support both `vol` and `volumes` — but I've only tested for the latter. What do you think about me removing `vol` support?

clee:
It lines up nicer with `env` but that’s literally the only reason I implemented it. I’m fine with you taking it out!

toolmantim:
Ok cool, thanks for the backstory.

The `env` is there because we have the same alias for the `env` in command steps in the pipeline.yml  — so you can copy+paste. But because there's no `vol` equivalent, we can probably remove it from here.

I'll do that now, and merge! 🎉

PR #150, opened 2018-06-03
Add `--workdir` support for run steps.

asford:
Adds support for the `--workdir` option for `docker-compose run` steps
via a `workdir` plugin option. This provides support for step-specific
working directories, which may depend on the build-time environment.
This is intended to support `run` steps within the build directory when
used in combination with appropriate volume mounts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,13 @@ A list of KEY=VALUE that are passed through as build arguments when image is bei
 
 A list of either KEY or KEY=VALUE that are passed through as environment variables to the container.
 
+### `workdir` (optional, run only)
+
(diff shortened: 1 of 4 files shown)


toolmantim:
Thanks, this PR looks great! One thing is I think the example might be too specific to your use case, and potentially a little confusing. Do you mind updating the PR to remove that paragraph from the docs change? And then we can 🚀🎉

asford:
Agreed, the paragraph isn't helpful without a larger example context. Trimmed in 84b2f68.

toolmantim:
Thanks @asford!

PR #152, opened 2018-06-06
Add support for running service with the --service-ports flag

ykifle:
Docker compose does to map service ports to the host when using the `run` command by default, even when it is configured in the docker-compose.yml file. You can use the `--service-ports` flag with the `run` command to enable mapping the ports to the host.

This adds a `service-ports` plugin option to allow using the `--service-ports` flag in the `run` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,10 @@ A list of KEY=VALUE that are passed through as build arguments when image is bei
 
 A list of either KEY or KEY=VALUE that are passed through as environment variables to the container.
 
+### `service-ports` (optional, run only)
+
+Map the services ports to the host via `docker-compose run --service-ports`.
+
 ### `pull-retries` (optional)
 
 A number of times to retry failed docker pull. Defaults to 0.
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -101,6 +101,11 @@ if [[ "$(plugin_read_config DEPENDENCIES "true")" == "false" ]] ; then
   run_params+=(--no-deps)
 fi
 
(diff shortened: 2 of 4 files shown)


lox:
Thanks @ykifle, happy to accept this, although I'd love to know the use case!

ykifle:
Sure @lox. I'm trying to setup Selenium testing using Buildkite, nightwatch and Saucelabs. We have a Buildkite agent running in an EC2 instance that uses docker compose to run a container that starts up the server I am testing in a background process and then runs the nightwatch tests.

Nightwatch sends Selenium commands to the Saucelabs CI service which opens up web browsers and tries to load the website. So Saucelabs needs to be able to reach the server running in docker on the Buildkite agent. So we need to bind the port the server is listening too, to the host.

At least I think that's how to get things working. Let me know if you know a better way to do this.

toolmantim:
Exposing ports is probably not something you'd normally want to do in a Buildkite environment! But I can definitely see the use cases. But no-one has a use for it at the moment, maybe we close this for now, and leave it for Google. Happy to look at adding it if someone finds it missing again though!

toolmantim:
(only because every arg option adds a slight bit of overhead in terms of compatibility with docker-compose binary versions)

toolmantim:
Thank you for taking the time to put together such a great PR though! Happy to re-open and get this updated if there's need for it.

stan-irl:
I want to run a local integ test in my CI that starts a local server and client in separate containers running the code thats in my PR. To do this i need to expose the listen port on the server

PR #153, opened 2018-06-07
add --no-ansi optional parameter

gugahoi:
This PR adds an option to run the docker compose plugin with `--no-ansi` flag enabled. This is needed because sometimes ansi enabled outputs mess up the step timing in Buildkite's UI which ends up showing the timing in the wrong steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,12 @@ If set to false, doesn't start linked services.
 
 The default is `true`.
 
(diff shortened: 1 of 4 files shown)


toolmantim:
Thanks @gugahoi for the great PR!

gugahoi:
Wow you guys are quick!

PR #156, opened 2018-06-12
Expand local volume paths

### LINKED ISSUE #155: Volume mapping does not work as intended
Since releasing `v2.4.0` with volume mapping in the plugin it seems it cannot map relative folders as docker-compose needs the full path to map via the `-v` flags. Unfortunately it seems like we cannot get the full path via `$(pwd)` in `pipeline.yml` so I am unsure how to make use of the flag if not for using data volumes. Below is an example using docker compose cli:

```bash
$ ls -l docs/
total 40
drwxr-xr-x 2 gus gus 4096 May 15 09:55 development
-rw-r--r-- 1 gus gus  551 Dec 27 15:21 images.md
-rw-r--r-- 1 gus gus  456 Dec 27 15:21 storybook.md
drwxr-xr-x 2 gus gus 4096 Jun  5 09:15 testing
drwxr-xr-x 2 gus gus 4096 Dec 27 15:21 tools
drwxr-xr-x 2 gus gus 4096 May 15 09:55 workflow

$ docker volume ls
DRIVER              VOLUME NAME

$ docker-compose run -v ./docs:/tmp/Downloads --entrypoint ls app -l /tmp/Downloads
total 0

$ docker volume ls
DRIVER              VOLUME NAME
local               bd6d29ba0873a698bb03ecd333aaa1bf76afd769440f4385f1e7def98872c80a
local               docs

$ docker-compose run -v $(pwd)/docs:/tmp/Downloads --entrypoint ls app -l /tmp/Downloads
total 40
drwxr-xr-x 2 node node 4096 May 14 23:55 development
-rw-r--r-- 1 node node  551 Dec 27 04:21 images.md
-rw-r--r-- 1 node node  456 Dec 27 04:21 storybook.md
drwxr-xr-x 2 node node 4096 Jun  4 23:15 testing
drwxr-xr-x 2 node node 4096 Dec 27 04:21 tools
drwxr-xr-x 2 node node 4096 May 14 23:55 workflow
```

How can I achieve relative folder volume mapping with the plugin?

toolmantim:
Fixes #155, and the following example from the documentation, by expanding `./` to `$PWD` at run-time, because docker-compose only does relative volumes paths in the YAML-style configuration:

```yml
steps:
  - command: generate-dist.sh
    artifact_paths: "dist/*"
    plugins:
      docker-compose#v2.4.0:
        run: app
        volumes:
          - "./dist:/app/dist"
```

### DIFF
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -40,3 +40,12 @@ check_linked_containers_and_save_logs() {
     fi
   done
 }
+
+# docker-compose's -v arguments don't do local path expansion like the .yml
+# versions do. So we add very simple support, for the common and basic case.
+# 
+# "./foo:/foo" => "/buildkite/builds/.../foo:/foo"
+expand_relative_volume_path() {
+  local path="$1"
+  echo "${path/.\//$PWD/}"
+}
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -92,7 +92,7 @@ done <<< "$(printf '%s\n%s' \
   "$(plugin_read_list ENVIRONMENT)")"
 
 while IFS=$'\n' read -r vol ; do
-  [[ -n "${vol:-}" ]] && run_params+=("-v" "${vol}")
+  [[ -n "${vol:-}" ]] && run_params+=("-v" "$(expand_relative_volume_path "$vol")")
 done <<< "$(plugin_read_list VOLUMES)"
 
 # Optionally disable allocating a TTY
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -350,7 +350,7 @@ export BUILDKITE_JOB_ID=1111
 
   stub docker-compose \
     "-f docker-compose.yml -p buildkite1111 -f docker-compose.buildkite-1-override.yml pull myservice : echo pulled myservice" \
-    "-f docker-compose.yml -p buildkite1111 -f docker-compose.buildkite-1-override.yml run --name buildkite1111_myservice_build_1 -v ./dist:/app/dist -v ./pkg:/app/pkg myservice pwd : echo ran myservice with volumes"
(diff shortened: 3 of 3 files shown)


toolmantim:
@gugahoi @clee how's this look?

lib/run.bash
```diff
@@ -40,3 +40,12 @@ check_linked_containers_and_save_logs() {
...
   done
 }
+
+# docker-compose's -v arguments don't do local path expansion like the .yml
+# versions do. So we add very simply support, for the common and basic case.
+# 
+# "./foo:/foo" => "/buildkite/builds/.../foo:/foo"
+expand_relative_volume_path() {
+  local path="$1"
+  echo "${path/.\//$PWD/}"
```
lox: 🤔How about something like:

```bash
if [[ -f $1 ]] ; then
  (cd $(dirname $1); pwd)
elsif [[ -d $1 ]] ; then
  (cd $1; pwd)
fi
```

lox: Oh, I just noticed that you were replacing it with `$PWD`. Why not just resolve the path?

toolmantim: It just seemed to be the simplest thing that works? Happy to full resolve it using the trick you did though! I'll change it now.

lox: Yeah, fair enough. I reckon the fact that environment variables get expanded by the run eval is probably a bug. It makes me uncomfortable.

clee: > Whereas with -v I think it actually creates the path on the host filesystem if it doesn't exist yet?

I can confirm this is definitely true. If the host's path does not exist before the call to `-v`, then docker-compose will create it.

I think this looks fine as-is.

toolmantim:
Gonna run with this as-is for now!

lox:
Yeah, soz, I meant to say I agreed!

toolmantim:
No worries @lox :)

PR #163, opened 2018-08-12
Support cache_from directive in composefile 2.2+

lox:
Based on discussion in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/commit/e454d2bef1498fc5020a9d4201b9986fd468386d#commitcomment-30002868, this allows `cache_from` usage in compose file versions 2.2, where previously we only allowed 3.2. 

/cc @bjeanes @dreyks


bjeanes:
Might want to update README too, unless you want to test it first?

PR #172, opened 2018-09-19
Move cleanup to a pre-exit hook to handle cancel

### LINKED ISSUE #169: Canceled jobs don't clean up networks
We've discovered that currently running jobs using this plugin, when canceled, clean up their containers but don't clean up the networks that were created.  They *are* cleaned up when the job runs successfully.

This has resulted in our build server repeatedly breaking due to no IP space being left.

lox:
Currently when a job is cancelled, it will kill the command hook and often docker-compose won't have time to cleanup. This moves the cancellation to a `pre-exit` hook to give it more time to run.

Closes #169.


robholland:
Would be great to get this merged, it's causing build failures when the networks pile up.

lox:
Sorry for the delay @robholland, was doing some thinking about whether there was anything else this could break, but let's just merge it and see!

PR #176, opened 2018-10-10
Push stderr to log file as well instead of outputting to terminal

GreyKn:


lox:
Good catch! Thanks.

toolmantim:
This was released in [v2.6.0 (🎄 Holiday Cheer)](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/releases/tag/v2.6.0) 🎉

PR #179, opened 2018-11-27
[bug] Fix custom image path when a build repository is defined

SpencerSharkey:
This issue affects the `build` step, specifically the path creation of an image to push when a repository and image name are specified.

## Example Config:
```json
{
    "plugins": {
        "docker-compose": {
            "build": "app",
            "config": "some/directory/docker-compose.yml",
            "image-repository": "gcr.io/project/repo",
            "image-name": "app:latest"
        }
    }
}
```

## Problem/Solution:
The resulting image path would end up malformed.

#### Before fix:
`${image_repository}:${image_name}`
`gcr.io/project/repo:app:latest`

#### After fix:
`${image_repository}/${image_name}`
`gcr.io/project/repo/app:latest`

### DIFF
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -31,7 +31,7 @@ for service_name in $(plugin_read_list BUILD) ; do
   service_idx=$((service_idx+1))
 
   if [[ -n "$image_repository" ]]; then
-    image_name="${image_repository}:${image_name}"
+    image_name="${image_repository}/${image_name}"
   fi
 
   build_images+=("$service_name" "$image_name")


toolmantim:
Oh dear, thanks for the bug report and fix! We'll look at getting this tested and merged. Thanks @SpencerSharkey

lox:
Apologies for the confusion! `image-name` is unfortunately named. It's actually used as the tag part of the image name:

> The name to use when tagging pre-built images. If multiple images are built in the build phase, you must provide an array of image names.

In your case, is there a specific reason you want `latest` as the tag? That will mean that your steps all build and consume the same image (`app:latest`). It needs to be unique to a given build. The default is `${BUILDKITE_PIPELINE_SLUG}-${service_name}-build-${BUILDKITE_BUILD_NUMBER}`, which is created in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/3438eb9a60f11add15134531fd0fd7205c51c928/lib/shared.bash#L164-L175.

lox:
I'm going to close this for now, as we haven't heard back! Let us know if we can help solve the underlying problem!

PR #181, opened 2018-12-12
Use `label` over `name` for build step config in examples

petervandoros:
I noticed that `name` was being used instead of `label` in the example pipeline snippets in the README.md file, so I thought I'd create a small PR to make them consistent with the command build step documentation.


toolmantim:
Thanks so much @petervandoros! 🙏🏼 A huge improvement.

PR #182, opened 2018-12-13
Support use aliases so network aliases can be used with the plugin

snmaynard:
I use network aliases during some of our pipelines and so the `docker-compose run` command needs to have the `--use-aliases` option so that they actually take effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,6 +392,12 @@ If set to false, disables the ansi output from containers.
 
 The default is `true`.
 
+### `use-aliases` (optional, run only)
+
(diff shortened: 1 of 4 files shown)


snmaynard:
Apologies - done! I just lifted the explanation from the help output from `docker-compose run --help` but can adjust if you feel its cryptic in that context.

lox:
Looks good, thanks @snmaynard!

toolmantim:
This was released in [v2.6.0 (🎄 Holiday Cheer)](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/releases/tag/v2.6.0) 🎉

PR #186, opened 2018-12-26
Add command parameter with support for array-based arguments

### LINKED ISSUE #59: Add support for multiple commands?
It looks like multiple commands aren't supported:
```
steps:
  - label: ':hammer: Test'
    command:
      - 'echo COMMAND_1'
      - 'echo COMMAND_2'
    plugins:
      docker-compose:
        run: tests
        config: docker-compose.buildkite.yml
```

I see that it tries to run: `docker-compose -f docker-compose.buildkite.yml -p buildkite82973f2986fe406c9a67a03338bca53e run tests echo COMMAND_1 echo COMMAND_2`

And my guess is that https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/hooks/commands/run.sh#L43 passes in $BUILDKITE_COMMAND as an array.

Would it be possible to have that line in `run.sh` iterate over each item of `$BUILDKITE_COMMAND` and run then sequentially?

As a workaround, I'm including the repeated step inside the first command (a shell script), but it'd be nice to specify multiple command to re-use e.g. a script that processes xunit.xml output files.

lox:
This is a port of the learnings from https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/68.

It changes the default behaviour from trying to parse the `$BUILDKITE_COMMAND` into tokens into defaulting to `/bin/sh -e -c "$BUILDKITE_COMMAND"`. This removes the need to try and tokenize the arguments, which has lead to a lot of historical bugs and confusion.

Using a shell is useful because often users assume that a shell exists, and use commands with shell-isms like:

* `command1 && command2`
* `command > outfile`
* `echo hello world`
* `command < myfile.txt`
* `command1; command2; command3`

Also, critically, a shell is needed for multi-line commands, which are currently unsupported (closes  #59).

To support images that don't have `/bin/sh` in them, or where the `entrypoint` is set such that passing a shell wouldn't make sense, a new config option of `shell` and `command` are added.

## New: `shell` config

The `shell` parameter allows for the customization of the shell vs the default:

```yaml
steps:
  - command: "npm run tests"
     plugins:
       - docker-compose:
           run: "windows-app"
           shell: ["CMD.EXE", "/C"]
```

or you can disable the shell to get the previous behaviour:

```yaml
steps:
  - command: "npm run tests"
     plugins:
       - docker-compose:
           run: "app"
           shell: false
```

## New: `command` config

The `command` config also allows for very specific parsing of arguments to pass directly to the `docker-compose` command. A shell is not inferred when the `command` config is used. 

```yaml
steps:
  - plugins:
       - docker-compose:
           run: "app"
           command: ["yarn", "run", "tests"]
```

## Breaking change: Images are assumed to have a shell

The plugin now assumes the container running your command has a shell accessible at `/bin/sh`.

* If your container doesn’t have a shell, you’ll need to install a shell by adding one to your `Dockerfile`, or switching to using the new `command` plugin option.
* If your container has a shell in a location other than `/bin/sh` use the new `shell` plugin option to customize the path.
* If your container has a custom entrypoint that expects the raw arguments, use the new `command` plugin option.

## Breaking change: Windows support 

A subsequent PR will add automatic Windows detection, but in the meantime to use this plugin on Windows images set the new plugin `shell` option to `["CMD.EXE","/C"]`.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,7 +40,7 @@ steps:
     plugins:
       ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         build: helloworld
-        image-repository: buildkiteci/docker-compose-buildkite-plugin
+        image-repository: ${DOCKER_REPO-buildkiteci/docker-compose-buildkite-plugin}
         config: tests/composefiles/docker-compose.v2.1.yml
 
   - wait
@@ -87,7 +87,7 @@ steps:
     plugins:
       ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         build: helloworld
-        image-repository: buildkiteci/docker-compose-buildkite-plugin
+        image-repository: ${DOCKER_REPO-buildkiteci/docker-compose-buildkite-plugin}
         image-name: llamas-build-${BUILDKITE_BUILD_NUMBER}
         config: tests/composefiles/docker-compose.v2.1.yml
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -45,19 +45,19 @@ steps:
 
   - wait
   - label: run after build with v2.0
-    command: /hello
     plugins:
       ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         run: helloworld
         config: tests/composefiles/docker-compose.v2.0.yml
+        command: /hello
 
   - wait
   - label: run after build with v2.1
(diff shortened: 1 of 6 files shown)


bjeanes:
Nice! FWIW this is how I originally _expected_ it would work, before discovering that it didn't.

lox:
Ok, I wrote my heart our in the PR. Edits / suggestions welcomed @toolmantim.

PR #188, opened 2019-01-14
Add cache-from-command plugin entry

elliott-davis:
This adds the ability to pull cache image(s) from a command rather than a
static string.

This is useful when determining the latest image from metadata other than
a docker tag like latest.

Signed-off-by: Elliott Davis <elliott@excellent.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,10 @@ This option can also be configured on the agent machine using the environment va
 
 A list of images to pull caches from in the format `service:index.docker.io/myorg/myrepo/myapp:tag` before building, ignoring any failures. If multiple images are listed for a service, the first one to successfully pull will be used. Requires docker-compose file version `3.2+`.
 
+### `cache-from-command` (optional, build only)
+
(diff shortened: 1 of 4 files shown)


lox:
Sorry for the slow reply @elliott-davis, it's an interesting idea, did you decide against it or just get fed up with waiting for us? 😓

elliott-davis:
Hey @lox! I don't mean this in a bad way but a bit of both 😂. I found a work-around to the need for this PR and since it sat for a long time without interaction I assumed it might be one of those rot-on-the vine PR's that gets closed in a year. Happy to re-open if it's interesting, but I also understand the Buildkite team is busy.

Thanks for the follow up!

toolmantim:
😅

What was the workaround?

It’s the first time we have a “run a command to get the value of this setting”, so it’s been a bit hard to figure out how it fits in.

elliott-davis:
The workaround was basically using docker tags correctly. The use-case for
this PR was querying artifactory to get the latest passing version of a
build pushed to artifactory.

Instead of running the query to get the latest version from artifactory, we
just only push a tag for unstable if the entire pipeline passes.

I actually modeled this idea off the monorepo diff plugin that uses the
value of a script to pull the commit sha to get the number of commits to
trigger against.

PR #192, opened 2019-01-29
Correctly handle . in docker compose service names.

robholland:
Prior to this fix a service with periods in such as "kafka-0.broker.kafka" may cause the plugin to mistakenly use the kafka service's image for other services which didn't specify an image.

docker-compose has supported periods since 2015: https://github.com/docker/compose/pull/1624.

### DIFF
--- a/lib/push.bash
+++ b/lib/push.bash
@@ -5,7 +5,7 @@ compose_image_for_service() {
   local image=""
 
   image=$(run_docker_compose config \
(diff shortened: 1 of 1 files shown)


lox:
Many thanks!

lox:
Unfortunately this broke with `grep: invalid character range`.

robholland:
Hmm, sorry about that. We've used my branch in production and it's run without error, so not sure why it's worked for us.

Issue #138, opened 2018-05-11
Don't dump all container logs to the main output


grrowl:
`docker-compose` logs are uploaded as artifacts (as `docker-compose-logs/<container id>.log` files), but it's also dumped en masse to the main log. It'd be great to have the `docker-compose-logs` artifacts be named (or suffixed with the name), and avoid having those logs dumped out into the main log.

I'm aware there may be issues with some logs not uploading, or possibly that naming the docker-compose logs might not be possible. Hoping filing this issue will open some discussion, since our integration tests tend to error in the main container (and hence in the main logs) but it's significantly harder to debug a >1MB log with irrelevant output from selenium, chrome, rails db:seed output, etc.

lox:
I think the fact that the main container output gets written as a log file and uploaded on error is probably a bug.

Issue #63, opened 2017-07-21
Build step is confusing


lox:
Having a build step without an `image-repository` step doesn't make a lot of sense. Based on discussion with @toolmantim it sounds like we should:

- Deprecate `build`
- Add a `prebuild` step that is what `build` was, but with `image-repository` mandatory
- Clarify in the docs how it should be used

jgmchan:
I ran into this issue yesterday as I got the impression from the doco that the `build` step could save some time on a local build by prebuilding the image and utilizing the docker cache.

We don't have a docker registry to push to and only 1 agent so I just used the `run` step to do the building and the next steps can use the built image from the cache.

lox:
@toolmantim just thinking out loud for the v3 syntax, what if we got rid of the implicit pull / push stuff and just made it more obvious when pull / push was happening?

It would make things so much less magical. 

```yaml
env:
  REGISTRY: "xxx.dkr.ecr.us-east-2.amazonaws.com/my-app"

steps:
  - label: ":docker: Prebuild"
    plugins:
      - docker-compose#v3.0.0:
          - build: 
              service: app
              cache-from:
                - ${REGISTRY}:latest-${BUILDKITE_BRANCH}
                - ${REGISTRY}:latest
          - push:
              - ${REGISTRY}:build-${BUILDKITE_BUILD_ID}
              - ${REGISTRY}:branch-${BUILDKITE_BRANCH}
              - ${REGISTRY}:latest

  - wait

  - label: "Tests"
    plugins:
      - docker-compose#v3.0.0:
          - pull:
            - ${REGISTRY}:build-${BUILDKITE_BUILD_ID}
          - run: 
              service: app
              image: ${REGISTRY}:build-${BUILDKITE_BUILD_ID}

```

toolmantim:
I like the build and push in the first step, but then the pull and run is gonna be an epic amount of copy and paste config if you have 5+ different steps that use that same pre-built image (I’m thinking of buildkite/buildkite as an example)

toolmantim:
I really like this though.

Reckon the “images” key is needed for push and pull? Can they be anything other than images?

The old syntax allowed you to configure a stack-level cache destination… with this one you’d need the env at the top of every pipeline.yml? 🤔 Maybe we could auto-pickup a repository env, that’s used as default for all push and pulls it doesn’t include a repo host in the path?

lox:
[...]

PR #196, opened 2019-02-06
Cleanup dependency and linked service output

lox:
This PR uses `docker-compose up -d --scale myservice=0` to scale up service dependencies before the actual service runs, which means we can do it in a different header to avoid the noise of all the docker pull and build operations of linked containers.

The end result is ~~much~~ relatively cleaner output. The is the before:

image

This is after:

image

This also closes https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/138.

### DIFF
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -12,6 +12,12 @@ service_name_cache_from_var() {
   echo "cache_from__${service_name//-/_}"
 }
 
+if [[ -z "$image_repository" ]] ; then
+  echo "+++ 🚨 Build step doesn't have an image-repository defined"
+  echo "Without a repository to push to using a prebuild doesn't do anything. The run step will build before run"
+  exit 1
+fi
+
 # Read any cache-from parameters provided and pull down those images first
 for line in $(plugin_read_list CACHE_FROM) ; do
   IFS=':' read -r -a tokens <<< "$line"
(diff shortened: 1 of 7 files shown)


toolmantim:
The new output is great!

What's the new `require-prebuild` key? I've lots of questions about it, but would be great to understand the context first.

lox:
> What's the new `require-prebuild` key? I've lots of questions about it, but would be great to understand the context first.

Damn, I was hoping you wouldn't notice that. 

It's basically just for the test suite to ensure that run steps fail if there isn't a prebuilt image. Found a whole lot of bugs and failures that were showing up green. 

I wasn't planning on documenting it 😅

lox:
I also added a warning for when you try and `build` without an `image-repository`:

image

Might need a copy tweak now I read it again.

commands/build.sh
```diff
@@ -12,6 +12,11 @@ service_name_cache_from_var() {
   echo "cache_from__${service_name//-/_}"
 }
 
+if [[ -z "$image_repository" ]] ; then
+  echo "+++ ⚠️ Build step doesn't have an image-repository defined"
```
toolmantim:
```suggestion
  echo "+++ ⚠️ Build step missing image-repository setting"
```

commands/build.sh
```diff
@@ -12,6 +12,11 @@ service_name_cache_from_var() {
...
+  echo "Without a repository to push to using a prebuild doesn't do anything. The run step will build before run"
```
toolmantim:
```suggestion
  echo "This build step has no image-repository set. Without an image-repository, the Docker image won't be pushed to a repository, and won't be automatically used by any run steps."
```

toolmantim:
Cool, if it's internal, no problem! 👌🏼

lox:
> We should probably just deprecate `build` without an `image-repository`, if it's more often an accident than on purpose?

@toolmantim yeah, I reckon we will in the next major version. See https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/63

PR #201, opened 2019-02-23
Add build aliases

lox:
Often docker-compose setups will have multiple variants of services with different dependencies. For example:

```yml
version: “3”
services:
  app: &app
    build: .
  app-with-mysql:
    <<: *app
    depends_on: [ mysql ]
  app-with-postgres:
    <<: *app
    depends_on: [ postgres ]
  mysql:
    image: mysql
  postgres:
    image: postgres
```

Currently this needs to be expressed in a `build` as:

```yaml
steps:
  - label: ":docker: Build"
    plugins:
      - docker-compose#v2.6.0:
          build: [ app, app-with-mysql, app-with-postgres ]
  - wait
  - label: ":mysql:"
    plugins:
      - docker-compose#v2.6.0:
          run: app-with-mysql
  - label: ":postgres:"
    plugins:
      - docker-compose#v2.6.0:
          run: app-with-postgres
```

This works, because it re-uses the docker cache, but it ends up running the docker-compose build and then subsequent pushes three times, which is slow. 

This new syntax allows for aliasing services, which will do a single push and use that image for any future step matching the service and alias names:

```yaml
steps:
  - label: ":docker: Build"
    plugins:
      - docker-compose#v2.6.0:
          build: app
          build-alias: [ app-with-mysql, app-with-postgres ]
  - wait
  - label: ":mysql:"
    plugins:
      - docker-compose#v2.6.0:
          run: app-with-mysql
  - label: ":postgres:"
    plugins:
      - docker-compose#v2.6.0:
          run: app-with-postgres
```

Feels @matthewd @toolmantim?

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,10 @@ This option can also be configured on the agent machine using the environment va
 
 The name to use when tagging pre-built images. If multiple images are built in the build phase, you must provide an array of image names.
 
+### `alias` (optional, build only)
+
+Other docker-compose services that should be aliased to the main service that was built. This is for when different docker-compose services share the same prebuilt image.
+
 ### `args` (optional, build only)
 
 A list of KEY=VALUE that are passed through as build arguments when image is being built.
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ This option can also be configured on the agent machine using the environment va
 
 The name to use when tagging pre-built images. If multiple images are built in the build phase, you must provide an array of image names.
 
-### `alias` (optional, build only)
+### `build-alias` (optional, build only)
 
 Other docker-compose services that should be aliased to the main service that was built. This is for when different docker-compose services share the same prebuilt image.
 
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -92,7 +92,7 @@ if [[ -n "$image_repository" ]] ; then
     set_prebuilt_image "${build_images[0]}" "${build_images[1]}"
 
     # set aliases
-    for service_alias in $(plugin_read_list ALIAS) ; do
+    for service_alias in $(plugin_read_list BUILD_ALIAS) ; do
       set_prebuilt_image "$service_alias" "${build_images[1]}"
     done
 
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -87,8 +87,16 @@ if [[ -n "$image_repository" ]] ; then
   echo "~~~ :docker: Pushing built images to $image_repository"
   retry "$push_retries" run_docker_compose -f "$override_file" push "${services[@]}"
 
+  # iterate over build images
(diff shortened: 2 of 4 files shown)


toolmantim:
This makes sense! What’s the docker-compose.yml look like in this example?

lox:
Something like https://github.com/jwswj/samson/blob/05e8463c40f940f54a4f2ff993378f14c98f6c4b/docker-compose.yml @toolmantim

lox:
We might need to call it `build-alias`, `use-aliases` is confusingly a docker-compose thing we support.

toolmantim:
Thanks! I just updated the PR description with a Docker Compose example.

What do you think about the plural `build-aliases`? 🤔 But does that follow our other params?

It strikes me that before we add another use of “build” perhaps we should take this opportunity to cut a v3, with a renamed `build` to `prebuild`, and update the `image-repository` option name too (so confusing) — to `prebuild-destination` perhaps?

I think those renames, plus this and #202, makes for a good major release. Two big figures and a syntax improvement, that are all related.

Happy to do that rename PR later today, before this merge? I’ll mark “build” as deprecated and throw an error.

lox:
Yeah, making a v3 makes sense to me @toolmantim.

toolmantim:
Maybe just merging and shipping this and #202 makes sense. I had a branch 75% done that did the renames, but sounds like some of the ideas in #63 are more the way to go if we’re gonna introduce a breaking change.

matthewd:
If the "same app image just with different dependencies" pattern is common, perhaps there'd be value in directly supporting a `depends_on` under a `run` step?

```yaml
steps:
  - label: ":docker: Build"
    plugins:
      - docker-compose#v2.6.0:
          build: app
  - wait
  - label: ":mysql:"
    plugins:
      - docker-compose#v2.6.0:
          run: app
          depends_on: mysql
  - label: ":postgres:"
    plugins:
      - docker-compose#v2.6.0:
          run: app
          depends_on: postgres
```

Especially as the dependency list got longer, maintaining extra compose entries for each needed combo just felt like indirection.

toolmantim:
I like that idea @matthew! Though I’m not entirely sure it is common.

#127 is another example that could take advantage of this new aliases functionality. I’ve been wanting to help out @KevinGrandon with a solution ever since he opened that PR.

lox:
> Maybe just merging and shipping this and #202 makes sense. I had a branch 75% done that did the renames, but sounds like some of the ideas in #63 are more the way to go if we’re gonna introduce a breaking change.

💯

ngan:
@toolmantim 
> the "same app image just with different dependencies" pattern is common

This is exactly our use case.

@lox can you clarify how do we use `build-alias` when we're building multiple images? eg https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#building-multiple-images

PR #202, opened 2019-02-23
Add skip checkout option

lox:
Often steps that use prebuilt images don't need to checkout the repo code. This allows the checkout to be skipped:

```yaml
steps:
  - label: ":docker: Build"
    plugins:
      - docker-compose#v2.6.0:
          build: app
  - wait
  - label: "Run"
    plugins:
      - docker-compose#v2.6.0:
          run: app
          skip-checkout: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,10 @@ Set the shell to use for the command. Set it to `false` to pass the command dire
 
 Example: `[ "powershell", "-Command" ]`
 
+### `skip-checkout` (optional, run only)
+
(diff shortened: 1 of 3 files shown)


lox:
Maybe `checkout: (true|false)` instead?

toolmantim:
(I preferred `skip-checkout` — it had a much narrower meaning, and is clearly an on/off type thing)

lox:
Cool cool.

PR #203, opened 2019-03-01
WIP: Add optional dependent services to start

elliott-davis:
I wanted to get this PR out early and get notes before I moved too far along with it.

The gist of it is, links in docker-compose are deprecated and there really isn't a clean way (that I've found) to start dependent services inside of compose unless you do it manually. That's what this PR adds, a deps array to specify dependent services in your yaml that will be started at a scale of 1 for you to test against.

I'm happy to take any feedback or questions you might have around motive or what the end result looks like.

Signed-off-by: Elliott Davis <elliott@excellent.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,10 @@ A list of services to push in the format `service:image:tag`. If an image has be
 
 Pull down multiple pre-built images. By default only the service that is being run will be pulled down, but this allows multiple images to be specified to handle prebuilt dependent images.
 
+### `deps` (optional, run only)
+
+Start dependent services for your service. By default only your service is started (and any accompanying linked services). However, since links is officially deprecated this command is useful for starting what would normally be linked services.
+
 ### `config` (optional)
 
 The file name of the Docker Compose configuration file to use. Can also be a list of filenames.
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -21,6 +21,7 @@ pull_params=()
 up_params=()
 pull_services=()
 prebuilt_candidates=("$run_service")
+service_deps=()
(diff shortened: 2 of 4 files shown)


toolmantim:
Have you investigated the Docker Compose `depends_on` configuration property? I think that does what you might need, and we may not need an extra plugin configuration option.

toolmantim:
Thank you for taking the time to submit a Pull request though! That’s awesome 🙌🏼😊

elliott-davis:
Afaik the depends_on syntax was removed in compose version 3. Links is
still there but is marked for deprecation. I am currently leveraging
external_links but those don't start dependent containers in compose.

lox:
> Afaik the depends_on syntax was removed in compose version 3

I don't believe so, that's basically the point of docker-compose! Check out https://docs.docker.com/compose/compose-file/#depends_on-1. 

They deprecated [links](https://docs.docker.com/compose/compose-file/#links) a while back. 

I'd be in favour of a `depends_on` param that allows extra dependencies to be added though!

elliott-davis:
> Version 3 no longer supports the condition form of depends_on

I misread that line in the docs

I'll get this PR fixed up based on your suggestion. Thanks again for the feedback!

toolmantim:
Alternatively, if you don't need the option anymore, it's probably best if we just leave it until someone needs the extra functionality?

elliott-davis:
depends on worked as advertised - closing this - thanks again @toolmantim

toolmantim:
🙌🏼 Nice one @elliott-davis

Issue #206, opened 2019-03-11
Option for upload all service logs


pecigonzalo:
The PR #196 introduced some other changes than the ones described but one of them I believe should be reverted.
```
service_exit_code="$(docker inspect --format='{{.State.ExitCode}}' "$service_container_id")"

# Capture logs if the linked container failed
if [[ "$service_exit_code" -ne 0 ]] ; then
```

Makes it so only the failed container logs are uploaded, which I don't believe is the desired default behavior. Most compose groups, might have a failure in 1 container, due to something that happened in another, that did not fail.
EG: You get a 401 in a curl test, but you actually need the NGINX logs to better understand what happened.

I would propose to add: https://github.com/pecigonzalo/docker-compose-buildkite-plugin/tree/feature/log-options

RootTJNII:
+1.  I have a failing test where microservices run in multiple containers and the plugin runs a test runner container.  There is a problem with one of the microservices but I now can't see the log.  I agree with @pecigonzalo that the logs of all the containers should be uploaded, for the same reason, and that this behavior change was unexpected and not desirable in our environment.

toote:
It would appear that PR #210 implemented the exact functionality necessary. Closing

Issue #205, opened 2019-03-06
Running migration steps or multiple compose


pecigonzalo:
Hi! we have some integration tests that require some database migrations, i wonder what is the recommended pattern for this (maybe outside this plugin scope).

I tried doing:
```
[...]
- label: ":hammer: Integration"
    plugins:
      - docker-compose#v2.1.0:
          run: migrator
          tty: false
          config:
            - docker-compose.yml
      - docker-compose#v2.1.0:
          run: integrator
          tty: false
          config:
            - docker-compose.yml
[...]
```
but since the images are cleaned up after the plugin runs, im left with no migrated DB.

We have a `depends'_on` in the compose file, but that is no guarantee the migration is actually finished.

I thought of using the integrator to query the DB or perform the migration, but to be honest, that seems rather awkward and not 1to1 on how the migration is run outside of Integration, not even close.

toolmantim:
Hi there! It's a good question 🤔

Could you provide a minimal docker-compose.yml along with those pipeline.yml steps, so we can reproduce it locally and think about what a solution might look like?

pecigonzalo:
Sure thing:
docker-compose.yml:
```
version: "2"

networks:
  app:
  db:

services:
  app:
    image: "nginx"
    depends_on:
      - migrator
      - db
    networks:
      - app
      - db

  migrator:
    image: "boxfuse/flyway:5-alpine"
    command:
      [
        "-url=jdbc:postgresql://db:5432/",
        "-user=postgres",
        "-password=mysecretpassword",
        "-connectRetries=60",
        "migrate",
      ]
    volumes:
      - ./migration:/flyway/sql
    depends_on:
      - db
    networks:
      - db

  db:
    image: postgres:9.6
    environment:
      POSTGRES_PASSWORD: mysecretpassword
    ports:
      - 5432
    networks:
      - db
```

docker-compose.it.yml:
```
version: "2"

services:
  integrator:
    image: "appropriate/curl"
    command: ["http://app/"]
    depends_on:
      - app
    networks:
      - app
```

migration/V1__create_users_table.sql:
```
create table users (
  id UUID primary key,
  name text not null,
  email varchar(100),
  dob date
);

```


We assume `app` connects to the DB and `integrator` performs a query that requires this app to connect to the DB.

Pipeline:
```
steps:
  - label: ":hammer: SBT Integration"
    plugins:
      - docker-compose#v2.1.0:
          run: migrator
          config:
[...]

PR #207, opened 2019-03-12
Add all-logs and rm options

pecigonzalo:
Related to: 
- #205 
- #206 
- #196

#### All Logs
As explained in #206, in many cases other container logs are required in order to properly troubleshoot a failed run.

#### `--rm` Option
Adding the `--rm` option and making it default to ensure you can run multiple runs of the same container or compose file.

#### Re-add linked container PS status
This allows for looking at the status before exit, I don't understand why it was removed in #196 as its quite useful for debugging a container being OOMd or similar.

## What
- Add `all-logs`
- Add `rm`
- Add linked container status output

### TODO
- [ ] Add tests
- [ ] Update docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,12 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
+### `all-logs` (optional)
+
+If set to false, only upload failed container logs.
+
+The default is `true`.
+
 ## Developing
 
 To run the tests:
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -131,6 +131,11 @@ if [[ "$(plugin_read_config USE_ALIASES "false")" == "true" ]]; then
   run_params+=(--use-aliases)
 fi
 
+# Optionally remove containers after run
+if [[ "$(plugin_read_config RM "true")" == "true" ]]; then
+  run_params+=(--rm)
+fi
+
 run_params+=("$run_service")
 
 if [[ "${BUILDKITE_PLUGIN_DOCKER_COMPOSE_REQUIRE_PREBUILD:-}" =~ ^(true|on|1)$ ]] && [[ ! -f "$override_file" ]]; then
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -23,36 +23,36 @@ pull_services=()
 prebuilt_candidates=("$run_service")
 
 # Build a list of services that need to be pulled down
-while read -r name ; do
-  if [[ -n "$name" ]] ; then
+while read -r name; do
+  if [[ -n "$name" ]]; then
     pull_services+=("$name")
 
-    if ! in_array "$name" "${prebuilt_candidates[@]}" ; then
+    if ! in_array "$name" "${prebuilt_candidates[@]}"; then
       prebuilt_candidates+=("$name")
     fi
   fi
-done <<< "$(plugin_read_list PULL)"
+done <<<"$(plugin_read_list PULL)"
 
 # A list of tuples of [service image cache_from] for build_image_override_file
 prebuilt_service_overrides=()
 prebuilt_services=()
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -417,6 +417,12 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
+### `all-logs` (optional)
+
+If set to false, only upload failed container logs.
+
+The default is `true`.
```
dan-kez: You probably want to add the rm option in the docs as well

pecigonzalo: Yep, it's on my TODO section. I wanted to check first if this was OK or not.

pecigonzalo:
Please let me know if you are OK with this changes, so I proceed with the TODO

toolmantim:
Sorry for the delay. I think the white space changes, and adding two options in the one pull request, was making it hard to figure out how to review this properly. Is there a chance you could do that? If not, it might have to wait until we do that for you and add any required changes.

pecigonzalo:
Yeah, I can split the changes into 3 PRs.

toolmantim:
Also I realized another thing that was making it tricky, not having any changes to the tests, which makes it hard to guess how it might affect existing users.

I think for both of these options we’ll want to keep our existing default bevhaioir. Is `rm` already true by default, but this gives you the option to disable it for your multiple runs?

pecigonzalo:
I did not do the test as I did not know if this would be even a desired feature for you, so I only made the "simplest" changes to showcase the features.

The code does not do `rm` at the moment (except at cleanup) but as far as I can tell, the behavior doesn't change since we dont pull the logs for the main container, as they are printed in the console.

The other changes, as explained basically revert partially the behavior changes done in #196, which did a bunch of unlisted major changes.
eg. https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/196/commits/63a9cc1c7f8804cd8455f4aa875a4f943c451ab8 which is a complete change in how commands are passed, was just basically "wrapped" in a "cleanup output" PR, with no explanation whatsoever of the impact (basically you cant pass `do this && that` or multiple commands anymore because its taken as a single arg, IMHO, not a great choice).

toolmantim:
I can understand the frustration of the changes not being well documented in the PR. Sorry about that! 

We’re happy to revisit changes, but they were made for good reasons and we need to think hard about tradeoffs — having clear before and after examples of how it broke things for you, and what subtleties these new PR changes might introduce on top of them, helps us all work together to design a way forward.

If you could split this into separate PRs with test changes (with no whitespace changes), that’d be much appreciated 🙏🏼 Then we could have better conversations and figure out the subtleties in each. And if not, that’s okay too, we’ll investigate getting all these issues looked at in good time.

pecigonzalo:
Yeah, no worries at all, I'll split into 3 PR and close this one after. I'm
not trying to "bash" on the other PR but rather explaining my reasoning
here.
Given there seems to interest in the changes in happy to write the test and
other missing parts, I just don't start with that in case the changes are
not desired and to avoid the overhead while just showing the general idea.

toolmantim:
I think both of these options make sense to add!

What about `upload-logs` with the valid values `”service”` (default), `”all”` and `false`? I think that matches current behavior, and gives you a new `upload-logs: “all”` option?

I’m trying to think through what `rm: true` would do, and what `rm: false` would do, and how it relates to current behavior. But let’s discuss in a new PR, so we don’t end up with the same confusing PR situation.

toolmantim:
Or perhaps `upload-logs` with `”service”` (default), `true` (upload all logs), `false` (upload no logs)

pecigonzalo:
Ill open 3 new PRs, and we can continue the discussion there.

pecigonzalo:
@toolmantim all PRs open

toolmantim:
Thanks @pecigonzalo! Sorry for the delay on the reviews 😓

PR #208, opened 2019-03-14
Added build-parallel build only option

hartror:
Give access to the `--parallel` option in docker-compose build to help run builds faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ Sets the build step to run with `--no-cache`, causing Docker Compose to not use 
 
 ### `build-parallel` (optional, build only)
 
-Set the build step to run with `--parallel`, causing Docker Compose to run builds in parallel.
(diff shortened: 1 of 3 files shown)


lox:
Huh, neat. I didn't know that was a thing! Sounds like a good default?

lox:
Looks like it was only added in 1.23 (https://github.com/docker/compose/pull/6115). 

Happy to just merge this as is until it's more common. Thanks for this @hartror.

lox:
Perhaps could you add a note in the docs that it needs docker-compose 1.23+?

hartror:
Yeah me neither but I thought I'd check before I refactored a pipeline to split the builds.

hartror:
Done.

hartror:
Now I better go update our agents to `1.23.1` 😏

PR #210, opened 2019-03-18
Add all-logs option

pecigonzalo:
Related to:
- #206
- #207 

#### All Logs
As explained in #206, in many cases other container logs are required in order to properly troubleshoot a failed run.

## What
- Add `all-logs`
- Restore upload all logs as the default behavior

### TODO
- [x] Add tests
- [x] Update docs

### Notes
Contains whitespace fixes, a product of `shfmt`. I would recommend we do enable and set it, so we stop having PRs from different people having different formats.

### DIFF
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -36,10 +36,11 @@ check_linked_containers_and_save_logs() {
   [[ -d "$logdir" ]] && rm -rf "$logdir"
   mkdir -p "$logdir"
 
+  # Get array of containers
   containers=$(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
-  IFS=$'\n'
+  IFS=$'\n' # Change IFS to new line
   for line in ${containers} ; do
-    if [[ -z "${line}" ]]; then
+    if [[ -z "${line}" ]] ; then
       # Skip empty lines
       continue
     fi
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -23,6 +23,15 @@ compose_cleanup() {
 check_linked_containers_and_save_logs() {
   local service="$1"
   local logdir="$2"
+  local uploadlogs="$3"
+  local uploadall="false"
+
+  if [[ "$uploadlogs" =~ ^(false|off|0|never)$ ]]; then
+    # Skip all if we are not uploading logs
+    return
+  elif [[ "$uploadlogs" =~ ^(true|on|1|always)$ ]]; then
+    uploadall="true"
+  fi
 
   [[ -d "$logdir" ]] && rm -rf "$logdir"
   mkdir -p "$logdir"
@@ -38,6 +47,7 @@ check_linked_containers_and_save_logs() {
     service_name="$(cut -d$'\t' -f2 <<<"$line")"
     service_container_id="$(cut -d$'\t' -f1 <<<"$line")"
 
+    # Skip uploading logs for the primary service container
     if [[ "$service_name" == "$service" ]] ; then
       continue
     fi
@@ -48,7 +58,9 @@ check_linked_containers_and_save_logs() {
     if [[ "$service_exit_code" -ne 0 ]] ; then
       echo "+++ :warning: Linked service $service_name exited with $service_exit_code"
       plugin_prompt_and_run docker logs --timestamps --tail 5 "$service_container_id"
-      docker logs -t "$service_container_id" &> "${logdir}/${service_name}.log"
+      docker logs -t "$service_container_id" &>"${logdir}/${service_name}.log"
+    elif $uploadall; then
+      docker logs -t "$service_container_id" &>"${logdir}/${service_name}.log"
     fi
   done
 }
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -38,6 +38,7 @@ check_linked_containers_and_save_logs() {
 
   # Get array of containers
   containers=$(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
+  OLDIFS="$IFS"
   IFS=$'\n' # Change IFS to new line
   for line in ${containers} ; do
     if [[ -z "${line}" ]] ; then
@@ -64,6 +65,7 @@ check_linked_containers_and_save_logs() {
       docker logs -t "$service_container_id" &>"${logdir}/${service_name}.log"
     fi
   done
+  OLDIFS="$IFS"
 }
 
 # docker-compose's -v arguments don't do local path expansion like the .yml
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -37,10 +37,8 @@ check_linked_containers_and_save_logs() {
   mkdir -p "$logdir"
 
   # Get array of containers
-  containers=$(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
-  OLDIFS="$IFS"
-  IFS=$'\n' # Change IFS to new line
-  for line in ${containers} ; do
+  mapfile -t containers < <(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
+  for line in "${containers[@]}" ; do
     if [[ -z "${line}" ]] ; then
       # Skip empty lines
       continue
@@ -65,7 +63,6 @@ check_linked_containers_and_save_logs() {
       docker logs -t "$service_container_id" &>"${logdir}/${service_name}.log"
     fi
   done
-  OLDIFS="$IFS"
 }
 
 # docker-compose's -v arguments don't do local path expansion like the .yml
--- a/README.md
+++ b/README.md
@@ -427,6 +427,16 @@ If set to true, docker compose will remove the primary container after run. Equi
 
 The default is `true`.
 
+### `upload-container-logs` (optional, run only)
+
+Select when to upload container logs.
+
+- `on-error` Upload logs for all containers when an error occurs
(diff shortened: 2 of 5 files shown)


pecigonzalo:
@toolmantim Im trying to write tests for this, but https://github.com/lox/bats-mock/blob/master/binstub#L4 eats my the `stdin` in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/210/files#diff-a8e4f810e317abc8d44cbecc28d7ab75L30 and make it impossible to write tests for. Any ideas?

toolmantim:
I think this one makes sense. @lox do you think it should upload all the container logs by default, or just the service log? I can see cases where you need the linekd container logs to debug a failure.

pecigonzalo:
To further add to this, the reason to make this default `true` is that when you have a failed build you probably do want all info as otherwise you also have to PR a change to enable logs. Its IMO the odd case that you want not to upload all logs.

pecigonzalo:
Removed all the spaces, but i can use some help with the test issue.

pecigonzalo:
Rebased changed

toolmantim:
I tried getting this to pass locally on my machine, but it's tricky. I couldn't quite work out the logic to be honest.

I don't think we want to be overriding `plugin_prompt_and_run` in the test? And I think we want the `elif $uploadall; then` as the first `if` statement, don't we?

Sorry for all the questions! I was hoping to just push a fix commit.

lox:
I'm ok with going back to "upload all logs on failure" as the default.

pecigonzalo:
> I tried getting this to pass locally on my machine, but it's tricky. I couldn't quite work out the logic to be honest.
> 
> I don't think we want to be overriding `plugin_prompt_and_run` in the test? 
Yeah it has not been easy :( . I stubbed the `plugin_prompt_and_run` to try and make the test simpler, as that section was not being tested, but rather the "do we also run `docker logs` for non failed containers"

> And I think we want the `elif $uploadall; then` as the first `if` statement, don't we?
I left the default order as to keep the functionality of lines https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/210/files#diff-a8e4f810e317abc8d44cbecc28d7ab75R56 and R57 which puts a nice and helpful stdout and warning in case of failure.

toolmantim:
Let’s remove the new `upload-logs` option and just make upload all logs if the main service fails?

toolmantim:
Maybe we could try an alternative approach with the tests, and just modify the existing suite? If we’re stuck on the tests.

pecigonzalo:
> Let’s remove the new `upload-logs` option and just make upload all logs if the main service fails?

I dont think this is the best approach, given a similar situation as in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/212 were some non primary container failed. I think we should either keep the option as is, or do upload all with non-zero exit code.
I personally prefer to be able to upload all logs, maybe we can tar/zip them all together so we dont upload 100 artifacts, but having all logs is super helpful for test that require multiple containers.

pecigonzalo:
Any further thoughts on this?

toolmantim:
You mentioned in the initial PR description that this was added in order to debug failure. I still think uploading logs in the case of any failure, a good idea. And I probably confused things by saying failure in the main service/container only.

What if we removed the option, and had it upload all container logs on any failure?

toolmantim:
Sorry for the delayed reply @pecigonzalo 😢

pecigonzalo:
No worries at all regarding the delay, as I updated the other PR, I checked this.

IMO Its valuable to have all logs in all cases, as it allows for better inspection. What are the concerns with that? or what behavior are we trying to avoid? 
If its the amount of artifacts, I can just zip them all together or similar.

lox:
My main objection here is adding more config and complexity. How about something like:

```
### `upload-container-logs` (optional, run only)
	
Whether to upload container logs from the docker-compose run.

  - `on-error` Upload logs for all containers when an error occurs.
  - `always` Upload logs all container logs
  - `never` Do not upload any logs
	
The default is `on-error`.
```

This keeps it simpler somewhat and provides a sensible default.

toolmantim:
Uploading all logs for every run, on the off chance that someone wants to inspect/debug a successful build, seems a waste. There are definitely use cases for needing the option, such as if it’s needed for compliance purposes (a record of all logs from the code that produced a build artifact), but let’s not have it upload logs for successful runs by default.

@lox I like that suggestion. Perhaps with `false` instead of never:

```
### `upload-container-logs` (optional, run only)
	
Whether to upload container logs from the docker-compose run.

  - `”on-error”` Upload logs for all containers when an error occurs.
  - `”always”` Upload logs all container logs
  - `false` Do not upload any logs
	
The default is `”on-error”`.
```

lox:
I'm not keen on `false` without `true`, or on `always` without `never`. If possible I think that things should either be <bool> or <string> rather than a mix.

toolmantim:
Sure thing! https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/210#issuecomment-487234292 It is! 👌🏼

If you’d like to update this PR to that API @pecigonzalo we’d love to accept it. Otherwise we’ll look at getting this fixed in an upcoming version. Thanks for the work you’ve put into this!

pecigonzalo:
OK, would it make sense to set `false = never` `true = always` `default = on-error` ?

pecigonzalo:
@toolmantim @lox everything updated to match the API as discussed and added some testing for it.

lib/run.bash
```diff
@@ -23,19 +23,34 @@ compose_cleanup() {
...
 
   [[ -d "$logdir" ]] && rm -rf "$logdir"
   mkdir -p "$logdir"
 
-  while read -r line ; do
-    if [[ -z "${line}" ]]; then
+  # Get array of containers
+  containers=$(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
+
+  IFS=$'\n'      # Change IFS to new line
```
pecigonzalo: I had to rework the `while` to this `for` with `IFS` to enable testing, otherwise `BATS` eats the `stdin` to the loop and basically kills the rest of the testing.

pecigonzalo:
Rebased

pecigonzalo:
@toolmantim could we merge this?

pecigonzalo:
Rebased and added `mapfile` instead of `IFS`

pecigonzalo:
@lox @toolmantim what is holding this at the moment? anything I can do to help move it forward?

toolmantim:
It's just waiting on a proper review from @lox.

thomasbarton:
Any progress on this?

lib/run.bash
```diff
@@ -23,21 +23,31 @@ compose_cleanup() {
...
 
   [[ -d "$logdir" ]] && rm -rf "$logdir"
   mkdir -p "$logdir"
 
-  containers=$(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
-  IFS=$'\n'
-  for line in ${containers} ; do
-    if [[ -z "${line}" ]]; then
+  # Get array of containers
+  mapfile -t containers < <(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
```
lox: This will be incompatible with Bash 3.x

lox:
I made one comment, will merge once it's bash 3.x compatible.

lib/run.bash
pecigonzalo: Is there a reason to keep bash 3.x compat? 4.x was released 10 years ago and no LTS os seems to use it.

As far as I could check quickly using docker:
- Alpine: 5.0
- Ubuntu 16.04: 4.3
- Debian 8: 4.3
- Debian 7: 4.2
- Centos 7: 4.2
- RHEL 6: 4.1

lox: Older macOS versions, we could certainly look at increasing it in future, but I'd prefer not to do that as part of this.

toolmantim: It's mostly OSX that causes this:

```
bash-3.2$ bash --version
bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

We've dropped Bash 3.x support in this plugin before, and it caused lots of problems for OSX users. When OSX bakes in Bash 4.x then we can ease up this restriction, but until then we'll have to stick to Bash 3.x'isms… sorry @pecigonzalo!

pecigonzalo: Ill have to revert back to while or loop and IFS then, ill do it as a separate commit and keep the history, as IMO we should use `mapfile` as its a much better string->list tool.

I can test if now it's testable altho, to be honest, I was not even aware the plugins had a requirement to be bash 3.x compatible and this was just using an unused feature.

Keeping macOS compat is prob quite hard and limiting as not only bash but many Linux tools do not exist or are different over there.

pecigonzalo: @lox would something like `IFS=$'\n' read -r -a containers <<< <(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')` work for bash3?

pecigonzalo: I think so, as its used in other files

pecigonzalo: Im back to BATS eating the output if I use read and while. Could you check my comment on: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/222#issuecomment-506295024 ?

jwhitcraft:
Any update on this, I could really use this feature.

lox:
I'm going to merge this and we'll bump the major version.

PR #211, opened 2019-03-18
Add --rm option as default for runs

pecigonzalo:
Related to: 
- #207 
- #205
- #196 

#### `--rm` Option
Adding the `--rm` option and making it default to ensure you can run multiple runs of the same container or compose file.

## What
- Add `rm` option as default

### TODO
- [x] Add tests
- [x] Update docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,12 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
+### `rm` (optional, run only)
+
+If set to true, docker compose will remove the primary container after run. Equivalent to `--rm` in docker-compose.
+
+The default is `true`.
+
 ## Developing
 
 To run the tests:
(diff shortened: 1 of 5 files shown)


toolmantim:
Thanks @pecigonzalo, PR looks great! I'd love to merge this, but could you please remove all the whitespace changes? Having such a large diff creates merge conflicts with other PRs, and makes it hard for for people who maintain forks. Thank you!

pecigonzalo:
I can remove them @toolmantim , but I believe we should remove those spaces, as they are corrected by the IMO standard tool https://github.com/mvdan/sh otherwise each file has its own "style".
I can also split that into another PR if you wish, let me know your thoughts. (same applies to my other 2 PRs)

toolmantim:
I understand you believe we should switch to that formatting. I'm not against it, but mostly I'm just trying to help get these merged for you first. Including project-wide formatting changes with a PR is always going to make it a little harder to accept.

pecigonzalo:
Fully understand that ill disable `shellfmt` here and we can do that in a separate PR.

pecigonzalo:
Removed `shellfmt` changes

toolmantim:
Thanks @pecigonzalo!

PR #214, opened 2019-04-10
Update documented version to 3.0.1

gerrywastaken:
The 3.0.0 update had a significant issue that would result in build
agents being in a non working state. This caused all future builds
on those agents to fail.

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/169

This commit updates the README so that other people don't end up using
the docker-compose-plugin version that has this issue.

Some generic terms so others can find this issue when they run into it:
`ERROR: for buildkite _1  Cannot start service : driver failed programming external connectivity on endpoint buildkite _1 Bind for 0.0.0.0: failed: port is already allocated docker-compose buildkite plugin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following pipeline will run `test.sh` inside a `app` service container using
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.0.0:
+      - docker-compose#v3.0.1:
           run: app
 ```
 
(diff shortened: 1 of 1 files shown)


toolmantim:
Thank you! Looks like [Boomper](https://github.com/apps/boomper-bot) didn't do the update for us. In the meantime, thanks for the manual fix, and I'll look at what's up with Boomper.

gerrywastaken:
Oh nice I didn't know about Boomper. Cheers for the tip. :)

PR #216, opened 2019-04-11
Check if prebuilt image exists before doing "get"

### LINKED ISSUE #215: Don't log FATAL when prebuilt image missing
Currently the plugin will try to get a prebuilt image from build for any pull or run services. If there is no pre-built image the `buildkite-agent meta-data get` command results in a FATAL error logged, however the error itself doesn't matter.

This causes confusion when debugging failed pipelines, as the FATAL error can be part of the normal flow.

Proposed fix is to use the `buildkite-agent meta-data exists` command before performing a `get`

pauldthomson:
Fixes #215 

Check for prebuilt image in meta data before attempting to get it. Stops FATAL error from appearing in build logs which led to confusion when debugging failed pipelines

### DIFF
--- a/lib/metadata.bash
+++ b/lib/metadata.bash
@@ -1,10 +1,19 @@
 #!/bin/bash
 
+# Check agent meta-data exists
+function plugin_check_metadata_exists() {
+  buildkite-agent meta-data exists "$1"
+}
+
 # Read agent metadata for the plugin
 function plugin_get_metadata() {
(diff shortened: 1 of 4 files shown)


toolmantim:
Thanks for the great PR @pauldthomson! That FATAL output has bugged me for a long while too… thanks for taking the time to looking into, and contributing a fix.

pauldthomson:
NP :)

PR #217, opened 2019-04-15
Remove old build containers before trying to bing up new ones

gerrywastaken:
The docker-compose-buildkite plugin had an error in older versions where
it would leave containers hanging around which then took up ports needed
on new builds. Every day somebody seems to push up an old build
accidentally, so instead I'm clearing out existing containers every
time there is a new build.

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -153,6 +153,12 @@ fi
 
 # Start up service dependencies in a different header to keep the main run with less noise
 if [[ "$(plugin_read_config DEPENDENCIES "true")" == "true" ]] ; then
+  # There was a bug with a previous version of this plugin
+  # This is a temporary work around to clear out containers that
+  # are hanging around from previous builds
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks for the PR @gerrywastaken, and sorry you hit the bug with the old version.

This plugin is designed to be safely run in parallel on the same machine (multiple agents per machine), and can't assume you're not running other Docker containers too.

I think this change just indiscriminately kills all docker containers on the host?

gerrywastaken:
Oh wow! @toolmantim I didn't mean to create this as a PR. I think I must muscle memoried the `hub pull request` alias and accidentally created this. This was just actually a temporary fix on our fork (we currently only have one agent) which I'll later revert. It's not something that I wanted merged into the main repo. Sorry about the confusion and I'll look into how I automatically created this. 😆

toolmantim:
Heh! All good @gerrywastaken. I've done that myself too 😁

PR #219, opened 2019-04-29
Ensure docker_compose_config_version only matches "version: ..."

foygl:
The version matching regex will match things like `blahblahversion: blah` as well as just `version: '3.7'`. This can result in an invalid compose file which has a multiple line version.

Adding the '^' will ensure that we match from the beginning of the line and so exclude anything that doesn't begin with `version:`.


lox:
Good find! Thanks @foygl!

Issue #220, opened 2019-05-01
Error Exit Status 1: "docker inspect" requires at least 1 argument.


williams4027:
This morning we ran into some errors on our CI builds after the commands run successfully:

`"docker inspect" requires at least 1 argument.`

Looking through the code, it seems that this is due to the new check that was added in `run.sh`. I have included a build log, but wasn't sure what was incorrect with our setup. We are also using the Buildkite cloud formation stack. This also only affected one of our pipelines, but we rolled back the versions on all the others to be safe. Any thoughts?

Solution:
For now we rolled back to v3.0.2.

[root-mobile-server-integration_build_21518_generating-coverage-report.log](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/files/3135216/root-mobile-server-integration_build_21518_generating-coverage-report.log)

dgarbus:
We are also seeing this today for the first time, but are using an older version of the compose plugin (2.6.0).

pecigonzalo:
@williams4027 This should be fixed now, could you verify?

williams4027:
Sorry for the delay, I just ran the build using `40424b7` on one of our CI pipelines and was successful. That looks to have fixed it. Thanks everyone for the quick response!

PR #221, opened 2019-05-02
Don't attempt to output containers when using 'rm'

ags:
https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/211 added a '--rm' to the docker-compose commands in addition to the container cleanup done in the pre-exit hook.

This appears to silently break the check for linked log file uploads as no containers will be returned by `docker ps`, and conflicts with the changes in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/212 to output failed containers for the same reason, though the later fails loudly as no containers are passed to `docker inspect` (the former uses a more graceful loop).

The test suite was also broken due to conflicts between these two changes.

To fix things this tests for the 'rm' configuration and doesn't attempt to output failed containers or upload logs when present. That's not ideal as it's a bit unexpected; it would probably be worth reconsidering whether 'rm' should be enabled by default, and implementing a more graceful version of the failed container output check.

See https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/220

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -8,6 +8,7 @@ run_service="$(plugin_read_config RUN)"
 container_name="$(docker_compose_project_name)_${run_service}_build_${BUILDKITE_BUILD_NUMBER}"
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 pull_retries="$(plugin_read_config PULL_RETRIES "0")"
+remove_containers_after_run="$(plugin_read_config RM "true")"
 
 expand_headers_on_error() {
   echo "^^^ +++"
@@ -135,7 +136,7 @@ if [[ "$(plugin_read_config USE_ALIASES "false")" == "true" ]] ; then
 fi
 
 # Optionally remove containers after run
-if [[ "$(plugin_read_config RM "true")" == "true" ]]; then
+if [[ "$remove_containers_after_run" == "true" ]]; then
   run_params+=(--rm)
 fi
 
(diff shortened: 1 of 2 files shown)


toolmantim:
Cc @pecigonzalo

pecigonzalo:
Nice catch, odd as I ran a branch that has both these things without this problem (minus some small fixed we did at the end). Let me investigate, as the `--rm` did not affect this in my case.

pecigonzalo:
@williams4027 @ags could you confirm your docker-compose only contains 1 container definition? I believe the problem is related to that. 

We should not skip collecting logs on `--rm` but rather split the inspect here: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/212/files#diff-9ae713e61df74a2c7813f885fca363aeR267 to do it only if `docker_ps_by_project -q` has containers (there are none if only 1 container ran)

williams4027:
@pecigonzalo Just to make sure I give you the right information, my docker-compose.buildkite.yml that was being used has multiple (5) services defined, but the build step in the pipeline only targeted a single one:

```
"docker-compose" => {
  "config" => "docker-compose.buildkite.yml",
  "run" => "app"
}
```

Logs:
```
  | Building Docker Compose Service: app | 0s
  | ⚠️ No pre-built image found from a previous 'build' step for this service and config file. Building image...
  | $ docker-compose -f docker-compose.buildkite.yml -p buildkite84539356acc44c3bb7c4848875eb9614 build --pull app
  | app uses an image, skipping
  |  
  | Starting dependencies | 1s
  | $ docker-compose -f docker-compose.buildkite.yml -p buildkite84539356acc44c3bb7c4848875eb9614 up -d --scale app=0 app
  | Creating network "buildkite84539356acc44c3bb7c4848875eb9614_default" with the default driver
```

pecigonzalo:
@williams4027 and does it have any `depends_on` or `linked` containers? as in, does it start any other container if you start it?

williams4027:
Correct, that one specifically did not, no. Only a volume mount.

But other build steps in the same pipeline used this plugin and target `appwithserver`, which has `links` to other services. Those steps with multiple containers starting up did not fail. So I believe the idea that this only affects when a single container is started up for a step is correct based on my experience.

pecigonzalo:
I think a change like so:
```
❯ git diff              
Alias tip: gd
diff --git i/commands/run.sh w/commands/run.sh
index ec42d46..1f529bd 100755
--- i/commands/run.sh
+++ w/commands/run.sh
@@ -274,11 +274,15 @@ fi
 if [[ -n "${BUILDKITE_AGENT_ACCESS_TOKEN:-}" ]] ; then
   if [[ "$(plugin_read_config CHECK_LINKED_CONTAINERS "true")" != "false" ]] ; then
 
+    containers=($(docker_ps_by_project -q))
+
     # Get list of failed containers
-    failed_containers=($(
-      docker inspect -f '{{if ne 0 .State.ExitCode}}{{.Name}}.{{.State.ExitCode}}{{ end }}' \
-      $(docker_ps_by_project -q)
-    ))
+    if [[ 0 != "${#containers[@]}" ]] ; then
+      failed_containers=($(
+        docker inspect -f '{{if ne 0 .State.ExitCode}}{{.Name}}.{{.State.ExitCode}}{{ end }}' \
+        ${containers}
+      ))
+    fi
 
     if [[ 0 != "${#failed_containers[@]}" ]] ; then
       echo "+++ :warning: Some containers had non-zero exit codes"
```

Should be enought to fix this issue

ags:
Right, that makes sense. I'll close this and leave you to fix that then @pecigonzalo.

PR #224, opened 2019-06-18
Allow `push` to be combined with `build` or `run`

matthewd:
There doesn't seem to be any technical reason we can't allow `build`+`push`, and it seems useful for some workflows where you build an image, and then want to store it with multiple names (for future cache-from use).

I've been successfully using this in my pipeline for a while now -- though that admittedly only proves my particular configuration works.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -12,36 +12,41 @@ commands=()
 
 [[ -n "$(plugin_read_list BUILD)" ]] && commands+=("BUILD")
 [[ -n "$(plugin_read_list RUN)" ]] && commands+=("RUN")
-[[ -n "$(plugin_read_list PUSH)" ]] && commands+=("PUSH")
 
-# Check we've only got one of BUILD, RUN and PUSH
+# Check we've only got one of BUILD or RUN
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks! At first glance I think this disables using `push` by itself? Perhaps not intentional?

matthewd:
Definitely not intentional if so, but I don't think it does.

toolmantim:
Oh, I see! Yep, using `push` by itself looks okay. That conditional code at the start of the hook is getting a bit gnarly to be able to understand at a glance.

lox:
Yeah, this actually seems pretty pragmatic to me, I was going to rewrite the plugin to achieve this 🤷🏼‍♂️Shall we ship it @toolmantim?

toolmantim:
Yeah, I reckon we should! Once we've updated the plugin.yml schema and readme, we should be good to go.

Do you have any suggestions for the readme @matthewd, given your use case and what you would have been searching for?

lox:
Ooops, I just saw the todos.

toolmantim:
Any chance you could look at the readme/plugin.yml too @matthewd?

PR #235, opened 2019-08-20
Allow docker container creation during run to also receive build options

derekju:
I found it handy that I could define a step to run a command on a docker container without the pre-requisite of having that container pre-built already. However, the underlying `docker-compose` doesn't receive the same support for the plugin options that are only limited to the `build` run type.

I specifically wanted the ability to specify `ARGS` so I lifted the code from the `build.sh` command and dropped it in to the `run` command. This also allows for the `--no-cache` and `--parallel` options to be passed on as well.

My bash-fu isn't that great so not sure if there is a better way to DRY this up instead of copying & pasting but wanted to see if this was an acceptable idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,15 +387,15 @@ Prevent the removal of volumes after the command has been run.
 
 The default is `false`.
 
-### `no-cache` (optional, build only)
+### `no-cache` (optional, build and run only)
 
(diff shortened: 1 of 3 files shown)


toolmantim:
Thanks @derekju, and sorry for the delay. I added some tests, and updated the readme, and this is good to go 👍🏼

PR #242, opened 2019-11-04
Ensure containers array is defined

### LINKED ISSUE #241: Error with #210
I had some pipelines that didn't specify the docker-compose plugin version (they do now). They started failing this morning and it appears related to #210. The error in the buildkite log is 

`[2019-11-01T12:11:56Z] /var/lib/buildkite-agent/plugins/github-com-buildkite-plugins-docker-compose-buildkite-plugin/hooks/../lib/run.bash: line 23: containers[@]: unbound variable`

Our bad for using `master` and I've since fixed that - just wanted to let you know. Thanks!

lox:
Closes #241

### DIFF
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -42,7 +42,8 @@ check_linked_containers_and_save_logs() {
     [[ -n "$container" ]] && containers+=("$container")
   done < <(docker_ps_by_project --format '{{.ID}}\t{{.Label "com.docker.compose.service"}}')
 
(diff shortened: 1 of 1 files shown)


supaspoida:
@lox Hi! We just got tripped up trying to use the `upload-container-logs` option with the latest release and noticed that this fix is not included. When we switched to point our pipeline steps to master all was well. Would it be possible to cut a new release with this change included? Thanks!

PR #243, opened 2019-11-07
add graceful shutdown

calvin-pietersen:
add GRACEFUL_SHUTDOWN to stop containers
add GRACEFUL_SHUTDOWN test


lox:
Interesting! Would love to know a bit more about what prompted this change @calvin-pietersen!

calvin-pietersen:
@lox Yes, we have a scenario where one of our dependant containers needs to output an artefact. Which can only be done by gracefully shutting down the container :(

jathu:
@toolmantim can we merge?

toolmantim:
Sorry for the delay @jathu! That's been merged, and a release will be out soon.

PR #244, opened 2019-11-11
Allow build and run in one step

DuBistKomisch:
Inspired by #224, this removes the final restriction on using build and run together.

The use case for this is that for simply running a script in docker with a shared cached build, having to have `docker-compose build` > `wait` > `docker-compose run` overcomplicates the pipeline unnecessarily. In particular it prevents running many similar steps in parallel because of the `wait`.

As far as I can see there's no reason this shouldn't work: `build` sets the magic meta-data, `run` then immediately reads it like normal. It's worked so far in our testing.

Probably need to update the README and plugin schema including the changes from #224, not sure if I'm the right person to do that though.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -12,14 +12,6 @@ commands=()
 
 [[ -n "$(plugin_read_list BUILD)" ]] && commands+=("BUILD")
 [[ -n "$(plugin_read_list RUN)" ]] && commands+=("RUN")
-
-# Check we've only got one of BUILD or RUN
-if [[ ${#commands[@]} -gt 1 ]] ; then
-  echo "+++ Docker Compose plugin error"
-  echo "Only one of build or run is supported. More than one was used."
-  exit 1
-fi
-
 [[ -n "$(plugin_read_list PUSH)" ]] && commands+=("PUSH")
 
 # Don't convert paths on gitbash on windows


DuBistKomisch:
Been using this a bit more, still working well.

The only thing I think could go wrong here is if there's overlapping options for build/run, but the only such one is `pull-retries` which is pretty redundant for `run` since you pull in `build` instead. Even in that case users can still use the old split steps if they really want to anyway.

DuBistKomisch:
Well, with the latest `depends_on` this is not quite as big a deal since you don't need to mess with `wait`. It's still nice to have half the steps of course though. :smile:

huonw:
Hi @DuBistKomisch, thanks for publishing this. We've been trying this at https://github.com/stellargraph/stellargraph/pull/674, but unfortunately it isn't working as we hope. We noticed that the `run.sh` file ends with an `exit`:

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/dd0a3f41f2f3d362e68bc9e0924bf718b8f6bff5/commands/run.sh#L308

and thus (because it is being sourced with `.`, instead of executed), execution always stops after the `run`, and never reaches the `push`. I think this can be fixed by changing the `exit` to a `return`: https://github.com/huonw/docker-compose-buildkite-plugin/commit/a18b2c39e0682cf806be8f121750047bc05f7e4e

DuBistKomisch:
@huonw glad someone is using it :smile:

Sounds like that's probably an oversight from #224, not really anything to do with my change. Looking at it though, you probably want to do:

```
if [[ $exitcode -ne 0 ]] ; then
  exit $exitcode
fi
```

I think returning it will just make it be ignored and not correctly fail the step?

DuBistKomisch:
Never mind, `hooks/command` is `set -e` so should fail the step. I'd suggest submitting your fix as a separate PR :+1:

huonw:
Ah, I'd missed that it was closer to #224 👍

PR at https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/260.

DuBistKomisch:
Is there anything else I can do to get this merged? We've been using this everywhere in our pipelines for over a year and it's been working fine, would be nice to switch back to the official plugin version.

talentprince:
hi, @toolmantim, is there any update of this change? we also want this change ~
it's like we will run a build service and then build the app based on the build output, can't use docker plugin to do the run because there are other dependencies of other images so that we can only do docker-compose.

I think there are also many cases that people want run then build + push~

Hope to get your reply, thanks.

jeremybumsted:
@DuBistKomisch apologies for getting to this PR so late! At this point, we'd love to have this change added. The only thing we would need is some tests added for the change, and then we can add it. I'm going to close this PR, but if you are still interested in getting this change added, please feel free to open up a new PR with those tests and we'll get it reviewed and merged!

DuBistKomisch:
@jeremybumsted no worries, this is actually my last week at the company using this, so I don't think I'll have time to write tests for this, hopefully someone else can adopt it!

jeremybumsted:
I'm closing this PR, but moved the changes to a new PR in #325 so that we can add tests and get this feature shipped :)

PR #248, opened 2019-11-27
Check service names in cache-from against the services being built

huonw:
The service names in the cache-from lines are used to place
`cache_from: ...` configurations into the appropriate place in the
modified docker-compose config. Without that `cache_from`, the pulled
image cannot be used for caching, and so all the effort there is
useless.

Previously, a configuration like:

    build:
    - foo
    cache_from:
    - foo-bar-baz:example.com/foo-bar-baz:latest

would download the image, but be unable to use it for caching
purposes. This patch makes the plugin flag this as a soft error,
alerting the user that the `foo-bar-baz` cache-from line doesn't
correspond to any of the services being built.

This came up for us when we had long image names, and
abbreviated service names, and we'd accidentally used the
image name before the colon. It took a while to work out
why the caching wasn't behaving as expected.

### DIFF
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -38,6 +38,7 @@ for line in $(plugin_read_list CACHE_FROM) ; do
     echo "    $line"
     echo "Service name '$service_name' is not one of the services being built:"
     echo "    ${build_services[*]}"
+    exit 1
   fi
 
   echo "~~~ :docker: Pulling cache image for $service_name"
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -17,6 +17,8 @@ if [[ -z "$image_repository" ]] ; then
   echo "This build step has no image-repository set. Without an image-repository, the Docker image won't be pushed to a repository, and won't be automatically used by any run steps."
 fi
 
+build_services=$(plugin_read_list BUILD)
+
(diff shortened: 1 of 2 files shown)


huonw:
Hi @lox, we'd appreciate if you could take a look at this when you have a moment. 😄

jayco:
Hey, @huonw thanks for investigating and sending through the pull request, sorry it's taken so long for us to check this out.

I think given that this is a configuration error I wonder if we should look at returning an error and failing fast on this?

huonw:
I thought about that, but decided that caching is generally an optimisation rather than something required for correctness, so it would be unfortunate to cause existing plugin-users' build to hard fail. However, I guess the fact that people typically depend on specific versions alleviates this: when one upgrades to a version containing this patch, these problems will be flagged as part of that build and can be fixed then.

Thus, I'll follow your suggestion and make this an error.

huonw:
Hi @jayco, I updated this per your suggestion. Could you take a second look? Thanks!

pzeballos:
Hi @huonw! Sorry for the late reply! Thanks for your contributions, but we are going to close this one. It's not something we want to add at this point. Thanks!

Issue #197, opened 2019-02-07
Running with arguments doesn't seem possible


viraptor:
I've got a number of steps which use the same generic entry in `docker-compose.yml`. They're configured using an argument. This works just fine with `docker-compose` itself:
```
APP_PATH="Pipeline/Envato.Pipeline.Recurly.Tests" docker-compose run run_tests
```
This works with the following services description:
```
  run_tests:
    build:
      context: .
      dockerfile: Dockerfile.dotnet_test
      args:
        - APP_PATH
```

Unfortunately, it doesn't seem like the plugin can support it. It allows arguments in the `build` stage, but not `run`.

And if I do both build and run in a single step:
```
      - docker-compose#v2.6.0:
          build: run_tests
          args:
            - APP_PATH=....
      - docker-compose#v2.6.0:
          run: run_tests
```
I get a failure, because:
```
No pre-built image found from a previous 'build' step for this service and config file. Building image...
```
Even though it's just finished building one successfully.

Am I missing something? Or is this not possible / broken?

lox:
Sorry you've run into an issue @viraptor, I'll have to test it out, using multiple invocations in the one step isn't something we do often, so it's possible there is a gotcha in there.

DuBistKomisch:
:+1: noticed this too. If there's no preceeding `build` step, the `run` step needs to build the service, but doesn't use the `args` configuration when doing so. I've worked around it by just shoving it in `docker-compose.yml` too, but it prevents being able to use values from the buildkite environment, so definitely a problem.

Also, I think the reason the build and then run thing didn't work is because `build` doesn't set the meta-data `run` uses to find the pre-built image unless you use a `image-repository` etc. This doesn't make sense to do anyway of course, should only need `run` if only using local cache.

DuBistKomisch:
Opened a PR to fix this: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/250

redthor:
I'm having the same issue. I notice the original PR #250 was merged and then reverted (#261) and now there's #268  (:pray:  :crossed_fingers: )

Until #268 is done, @DuBistKomisch how do you 

>  just shoving it in `docker-compose.yml`

?

I'm trying to do that with variable substitution but having some difficulty.

e.g. I've tried passing both `args` and `env` from the pipeline:
```pipeline.yml
steps:
  - label: ":go: :gem: Linting check"
    plugins:
[...]

PR #250, opened 2019-12-06
Accept args and no-cache for run

DuBistKomisch:
Fixes https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/197, and threw in `no-cache` while I was at it since it's theoretically useful, but left out `build-parallel` since `run` only accepts one service anyway.

Wrote some tests here but also tested with our project and it's working.

Let me know if I've missed anything I need to change, I just grepped for the configuration property names.

`build_params` construction could probably be a shared method but oh well.


plugin.yml
```diff
@@ -89,10 +89,11 @@ configuration:
...
     leave-volumes: [ run ]
-    no-cache: [ build ]
+    no-cache: [ build, run ]
     use-aliases: [ run ]
     dependencies: [ run ]
     ansi: [ run ]
     tty: [ run ]
     workdir: [ run ]
     user: [ run ]
+    args: [ build, run ]
```
DuBistKomisch: this didn't exist and these don't seem to be in any particular order, so I just chucked it on the end here

jayco:
Hey, @DuBistKomisch thanks for sending this through, sorry it has taken a while to take a look at this.

As per https://github.com/buildkite/agent/pull/1161 the current agent version won't run multiple docker-compose plugin invocations in a single step. We're investigating that at the moment and once that's resolved we will come back to get this. Thanks for your patience!

DuBistKomisch:
@jayco I'm sorry, but I don't see how that relates to this PR?

jayco:
@DuBistKomisch Oh my mistake, I was looking at the example in the original issue and thought this might run into issues with multiple docker-compose invocations. This looks good to merge!

jayco:
@DuBistKomisch sorry we had to revert this we will get this fixed and merged.

DuBistKomisch:
no worries :)

PR #251, opened 2019-12-12
make clear -no-cache is boolean

TheDen:
Updates the docs so that it's clear for users that `-no-cache` is a boolean that can be set to `true` or `false`


toolmantim:
Thanks @TheDen!

PR #252, opened 2019-12-16
skip pulling cached images if no-cache is set to true

TheDen:
Updates the script such that If `no-cache` is set to true, `--cache-from` images are not pulled


toolmantim:
Hi @TheDen! Thanks for the contribution, and sorry it's taken a while to review. This change looks good… any chance you could add a test before we merge?

PR #254, opened 2020-01-09
Add propagate-uid-gid option

caiokf:
Similarly to what `docker-plugin` has, this PR is to add an option to choose whether to match the user ID and group ID for the container user to the user ID and group ID for the host user. It is similar to specifying user: 1000:1000, except it avoids hardcoding a particular user/group ID.

Using this option ensures that any files created on shared mounts from within the container will be accessible to the host user. It is otherwise common to accidentally create root-owned files that Buildkite will be unable to remove, since containers by default run as the root user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,12 @@ Specify the container working directory via `docker-compose run --workdir`.
 
 Run as specified username or uid via `docker-compose run --user`.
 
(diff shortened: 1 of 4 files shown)


jayco:
Thanks for taking the time to send this in @caiokf we appreciate the effort in adding this functionality.

PR #257, opened 2020-01-23
Fix expand_relative_volume_path on windows

### LINKED ISSUE #249: expand_relative_volume_path doesn't work for windows containers
Since it uses `$PWD` inside the hook it ends up with an emulated path like `/c/.../repo` which doesn't work when passed to `-v`, it has to be a native windows path like `C:\...\repo`.

Of course you can just use a relative path in `docker-compose.yml` instead but it's a bit inconsistent, so might be worth mentioning in the `volumes` docs at least?

DuBistKomisch:
Fixes #249.

I've tested this with a Windows container using a configuration like:

```
          volumes:
            - ./artifacts:C:\\workspace\\artifacts
```

I believe Linux containers on Windows also need the host path to be in native format, so it shouldn't break anything there, but I haven't personally tried that.

### DIFF
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -72,5 +72,14 @@ check_linked_containers_and_save_logs() {
 # "./foo:/foo" => "/buildkite/builds/.../foo:/foo"
 expand_relative_volume_path() {
   local path="$1"
-  echo "${path/.\//$PWD/}"
+  local pwd="$PWD"
+
+  # docker-compose's -v expects native paths on windows, so convert back.
+  #
+  # "/c/Users/..." => "C:\Users\..."
+  if is_windows ; then
+    pwd="$(cygpath -w "$PWD")"
+  fi
+
+  echo "${path/.\//$pwd/}"
 }


jayco:
Hey, @DuBistKomisch thanks for sending this through. 

This looks like a good fix, but having a look at this I noticed that you have used `cygpath`. Does that mean that we are assuming the use of `cygwin`?  We don't currently have a dependency on `cygwin` and we would be hesitant to add it.

DuBistKomisch:
@jayco good point, I probably should have mentioned this in the OP, since I did investigate it.

Firstly, cygpath isn't actually cygwin specific, msys/msys2 also include it since they descend from cygwin. Since Git for Windows is actually just bundling msys2 stuff, that means it also includes cygpath.

Secondly I believe since this plugin requires bash, it therefore requires Git for Windows as mentioned here: https://buildkite.com/docs/agent/v3/windows#git-for-windows. So really you do already have a dependency on this stuff, unless I'm missing another way to use this plugin on Windows.

jayco:
@DuBistKomisch thanks for the info, that makes sense! Happy to merge 🎉

DuBistKomisch:
thanks :smile:

Issue #259, opened 2020-02-05
Run followed by push never runs the push


huonw:
#224 allowed `push` and `run` to happen in a single step, but it seems that this doesn't actually work as intended because `run.sh` unconditionally `exit`s at the end:

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/94d9fcab9108f4d1e3b9ccce0e220f0d7b9a0184/commands/run.sh#L308

and so the shell running `command` hook will exit, because `run.sh`  is sourced in:

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/94d9fcab9108f4d1e3b9ccce0e220f0d7b9a0184/hooks/command#L45

PR #260, opened 2020-02-05
Fix run+push never executing the push stage

huonw:
In #224 we added the ability to do a build/run + push in a single step, but unfortunately there was a bug in the `run` command (the use of `exit`) that prevents the `build` stage ever being run. 

This PR replaces the `exit` with a `return`, which allows a successful `run` to continue on to `build`.

Fixes: #259

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -305,4 +305,4 @@ if [[ -n "${BUILDKITE_AGENT_ACCESS_TOKEN:-}" ]] ; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks so much for the investigation and pull request @huonw! Looks great. I just did some updates to the pull request to make it clearer what functionality this is fixing, for anyone wanting to upgrade or anyone looking for this fix.

huonw:
Thanks for merging!

PR #262, opened 2020-04-20
Add entrypoint option

### LINKED ISSUE #258: Clear entrypoint of service when creating docker-compose-override.yml
Images I work with often have an `entrypoint` set to run their default actions. When testing with this plugin, I typically need to create a custom `docker-compose-buildkite.yml` that sets `entrypoint: ""` so commands passed in my pipeline work correctly.

It would be really useful and hopefully not too complicated to add `entrypoint: ""` to the `docker-compose-override.yml` that this plugin creates, or alternatively support setting `entrypoint` in the plugin config for a `run` step.

emilio-notable:
Fixes #258 

The idea is to add an `entrypoint` option so we can override the one specified in the docker-compose file.

I don't have a lot of experience with docker or buildkite, so I'm not sure if this is the right way to do it. Please let me know if i missed anything, or if there is a better way to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,6 +435,10 @@ If set to true, docker compose will remove the primary container after run. Equi
 
 The default is `true`.
 
+### `entrypoint` (optional, run only)
+
+Sets the `--entrypoint` argument when running `docker-compose`.
+
 ### `upload-container-logs` (optional, run only)
 
 Select when to upload container logs.
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -150,6 +150,11 @@ if [[ "$(plugin_read_config RM "true")" == "true" ]]; then
   run_params+=(--rm)
 fi
 
+# Optionally sets --entrypoint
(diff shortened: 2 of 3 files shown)


scream314:
LGTM

toolmantim:
Thanks @emilio-notable! In general we hesitate to add options that can be set in the `docker-compose.yml` file itself — was there a use case that prevented you from doing that?

emilio-notable:
Hi @toolmantim,

our use case is that we have one `docker-compose.yml` that we would like to use for two buildkite steps, and the only difference should be the entrypoint.

right now we solved it using another docker-compose file, and specifying both in the buildkite step, but we were wondering if there was a cleaner way to do it, overriding just the entrypoint without having to create another docker-compose file

Thanks!

toolmantim:
Thanks for the details! That totally makes sense, and given the small footprint of the addition I think it’s totally reasonable, even if it’s something that can be achieved with separate docker compose config files.

Issue #240, opened 2019-10-26
Simplifying Plugin Configurration


elijahchancey:
Here's my current docker compose plugin configuration:
```
  - name: "Build"
    plugins:
      docker-compose#v3.1.0:
        build: "myapp"
        cache-from: myapp:111111111111.dkr.ecr.us-west-1.amazonaws.com/myapp:latest
        config:
          - docker-compose-buildkite.yml
        image-repository: 111111111111.dkr.ecr.us-west-1.amazonaws.com/myapp
        image-name: latest
        push:
          - myapp:111111111111.dkr.ecr.us-west-1.amazonaws.com/myapp:bk-$BUILDKITE_BUILD_NUMBER
```

I think it would've been easier to figure out if the plugin would have supported something like this:
```
  - name: "Build"
    plugins:
      docker-compose#v3.1.0:
        build: "myapp"
        cache-from-tag: 'latest'
        config:
          - docker-compose-buildkite.yml
        image-repository: 111111111111.dkr.ecr.us-west-1.amazonaws.com/myapp
        image-name: ['latest', 'bk-$BUILDKITE_BUILD_NUMBER']
```

There may be gotchas I'm not aware of...but it did take a bit too long for me to figure out my current (and working) configuration. Thanks!

lox:
Thanks for the context @elijahchancey. I'm confused what the `image-name: latest` is trying to do here, I think that might be creating a race condition for later steps 🤔

Totally hear you on making it easier to have the docker-compose plugin work out of the box with build and `cache-from` though!

elijahchancey:
When I push an image to ECR, it needs to be tagged both 'latest' and 'bk-123'. When I push a new image to ECR, it needs to be tagged 'latest' and 'bk-124'. At this point, the 'latest' will be removed from the image that is also tagged 'bk-123'. 

I need this functionality because I have a separate pipeline/api that needs to use myapp:latest to run tests.

toote:
I agree with the spirit of this issue, but the amount of scenarios currently covered by the plugin make it quite difficult to make such radical changes to it. We have made a note about them to start slowly implementing configuration (and documentation) changes that should make the plugin easier and more intuitive to configure in the future.

With all that said, I don't think it makes sense to have an issue open with no actual issue or objective finish line to achieve so I am closing this for the time being. If you disagree, feel free to leave a comment or open a new issue for us to discuss :)

PR #272, opened 2020-05-14
added the build option for dependency build

chetushivanna:
As part of the issue opened


toolmantim:
Fixes #240

toolmantim:
Do you have a sense if this is safe to add without any option @chetushivanna? Might it cause unnecessary building in places?

PR #279, opened 2020-07-04
Fix builds in pull requests from forks

toolmantim:
This makes the self-testing plugin syntax work with pull requests, by attempting to use the pull request repo if it's set, otherwise it falls back to using the pipeline's main repo.

Helps to fix tests in #278


toolmantim:
It appears there's another flakey test, but I'll merge this fix because it seems unrelated.

PR #281, opened 2020-07-15
Set default Buildkit output mode to `plain`

### LINKED ISSUE #256: Support for Buildkit?
Hey @lox, I've been looking around and I know there's no official support for buildkit baked into docker-compose as of now, but I found stumbled across https://github.com/lox/docker-compose-buildkit 😄

Are there any plans to use your shim for `docker-compose` in this plugin? It would be really useful! 

The most important feature I'm looking for is around `docker build --ssh`, since builds may reference private GitHub repos for dependencies.

In the meantime, would a decent approach be to have a step to manually pre-build an image with a `docker build && docker push` and have a `docker-compose.yml` that references that image directly for this plugin to use?

toolmantim:
As mentioned in #256, if you use [Buildkit](https://docs.docker.com/develop/develop-images/build_enhancements/) the default output can create a bunch of erroneous [log output groups](https://buildkite.com/docs/pipelines/managing-log-output#collapsing-output).

This change sets the default Buildkit output to `plain` to avoid this.

Fixes #256

### DIFF
--- a/hooks/pre-command
+++ b/hooks/pre-command
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -ueo pipefail
+
+# If they're using Buildkit, make sure it doesn't output the '---' lines that
+# mess up Buildkite's log output grouping
+export BUILDKIT_PROGRESS=plain
--- a/tests/pre-command.bats
+++ b/tests/pre-command.bats
@@ -1,0 +1,13 @@
+#!/usr/bin/env bats
+
+load '/usr/local/lib/bats/load.bash'
+
+@test "Pre-command sets Buildkit to use plain output" {
+  BUILDKIT_PROGRESS=
+
+  source "$PWD/hooks/pre-command"
+
+  assert_equal "plain" "${BUILDKIT_PROGRESS}"
+
+  unset BUILDKIT_PROGRESS
+}


toolmantim:
One thing I didn't consider, is that people still need to set `DOCKER_BUILDKIT=1` environment variable?

Which means, either:

1. We could just add a `buildkit: true` option to the plugin, which sets both of the environment variables for them
2. We could leave it as-is, but only set `BUILDKIT_PROGRESS` conditionally if `DOCKER_BUILKIT == "1"`

I think I'd prefer version 1. What do you think @jayco & @niceking?

niceking:
Hey @toolmantim! Totally keen to pair on this!!

niceking:
> One thing I didn't consider, is that people still need to set `DOCKER_BUILDKIT=1` environment variable?
> 
> Which means, either:
> 
> 1. We could just add a `buildkit: true` option to the plugin, which sets both of the environment variables for them
> 2. We could leave it as-is, but only set `BUILDKIT_PROGRESS` conditionally if `DOCKER_BUILKIT == "1"`
> 
> I think I'd prefer version 1. What do you think @jayco & @niceking?

I prefer 1 too! Nicer to be explicit rather than doing sneaky we-assumed-this-because-you-did-this things that we may forget about later!

toolmantim:
Another thing to note, is that it turns out there's a little more to this than we thought — there's also parts of the output that we'll need to modify if `buildkit` is set to true, it's not just about setting the output mode to plain.

toolmantim:
Turns out this doesn't actually solve the problem people were having after all, so let's close this for now.

PR #282, opened 2020-07-15
Fix --no-ansi option for docker-compose run

lucy-lufei:
The `--no-ansi` option is an option for `docker-compose`, so it should come before `run`, otherwise the command is not valid.


toolmantim:
Oh whoops, good catch @lucy-lufei, and thanks so much for the pull request!

PR #283, opened 2020-07-28
Fix entrypoint param

emilio-notable:
I think i made a mistake in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/262 😞  , which was causing the entrypoint option to be wrapped in quotes and not interpreted correctly by docker-compose. Sorry about that!

I think it should work now, i was able to test it with one of my BK pipeline using this branch (should have done that with my previous PR)


toolmantim:
Thanks @emilio-notable!

PR #286, opened 2020-09-18
Add ssh-agent support

halkeye:
Copied from docker plugin
I tried to make a test, but checking sockets is hard

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -8,6 +8,7 @@ run_service="$(plugin_read_config RUN)"
 container_name="$(docker_compose_project_name)_${run_service}_build_${BUILDKITE_BUILD_NUMBER}"
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 pull_retries="$(plugin_read_config PULL_RETRIES "0")"
+mount_ssh_agent=''
 
 expand_headers_on_error() {
   echo "^^^ +++"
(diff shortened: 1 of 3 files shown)


commands/run.sh
```diff
@@ -95,6 +96,25 @@ while IFS=$'\n' read -r vol ; do
...
+  if [[ -z "${IS_IN_TEST:-}" ]]; then
+    if [[ ! -S "${SSH_AUTH_SOCK}" ]] ; then
+      echo "+++ 🚨 The file at ${SSH_AUTH_SOCK} isn't a socket, has ssh-agent started?"
+      exit 1
+    fi
+  fi
+  run_params+=(
+    "-e" "SSH_AUTH_SOCK=/ssh-agent"
+    "--volume" "${SSH_AUTH_SOCK}:/ssh-agent"
+    "--volume" "${HOME}/.ssh/known_hosts:/root/.ssh/known_hosts"
```
vgrigoruk, line 114: I guess this should take https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#user-optional-run-only into account (do not assume the user is always `root`).

halkeye: > I guess this should take https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#user-optional-run-only into account (do not assume the user is always `root`).

your welcome to make the change. I'm no longer at the company that uses buildkite, so I'm not going to spend time on the fix sorry.

pzeballos:
Hi folks! Thanks for your contributions! You are welcome on submitting a new PR if you are still interested in the functionality. I'll close this one as it has conflicts and needs some changes.

PR #288, opened 2020-10-04
containers should stop on pre-artifact in case they generate artifacts on SIGINT

### LINKED ISSUE #267: compose_cleanup should run on pre-artifact hook
Some sidecar containers, like a vnc-recorder, generate artifacts on INT signal. However, docker compose cleanup is running after buildkite upload job artifacts.

I think we should move compose_cleanup from pre-exit hook to pre-artifact.

Let me know if that makes sense.

arturopie:
Fixes #267


arturopie:
Is there anything blocking this PR? Let me know if I forgot anything else.
Also, if it looks good, do you mind adding the `hacktoberfest-accepted` label to the PR?

lox:
Makes sense to me, I wonder if this will break backwards compatibility for anyone 🤔

arturopie:
> Makes sense to me, I wonder if this will break backwards compatibility for anyone 🤔

I don't see how yet, but my knowledge on the internals of this plugin is limited

jradtilbrook:
Hi @arturopie thanks for making this contribution. We've finally managed to get around to checking this out properly! ⌛ 

I've tested this and unfortunately, it introduces a breaking change. The [`pre-artifact`](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks) hook only runs if an artifact path is defined.

So this change will make the cleanup only happen if there are also artifacts to collect which won't suit every use case. I've shown this in the below pipelines you can check out:

1. [This pipeline](https://buildkite.com/jradtilbrook/docker-compose-pr-288/builds/3) shows the PR change compared to the current version when artifacts are defined: you can see they make the same logs and operate as expected.
Screen Shot 2022-02-02 at 09 36 29

2. [This pipeline](https://buildkite.com/jradtilbrook/docker-compose-pr-288/builds/4) shows what happens with your change when artifacts are not defined: the cleanup does not occur which is the issue here.
Screen Shot 2022-02-02 at 09 36 43


I'd say this can probably be worked around by perhaps having both hook scripts defined or doing some extra checking but it's a little messy. Are there any other methods to achieve your desired outcome? We'd love to have this PR updated so it doesn't break compatibility if you still need it. Otherwise please close this off if you no longer need the functionality.

jradtilbrook (changes requested):
Moving the hook to `pre-artifact` means that it will only run if artifact patterns are defined and so will break for other use cases. We need a different method or to make sure the hook still runs without artifacts.

arturopie:
Thanks for the explanation @jradtilbrook. It has been a while since I opened this issue, so I don't remember all the details. I think I ended up implementing a work around and move on. You can close this PR if you don't think it's useful anymore.

jradtilbrook:
No worries thanks for the contribution anyway @arturopie. We will close this

PR #307, opened 2021-08-03
Support omitting compose file version

vanstee:
As of docker-compose v1.27.0, the compose file version is now optional, defaulting to the [Compose Specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md), rather than v1.0. Here's the exact working from the [specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element):

> **Version top-level element**
> Top-level version property is defined by the specification for backward compatibility but is only informative.

and another mention that it's now optional in [their docs](https://docs.docker.com/compose/):

> A docker-compose.yml looks like this:
```
version: "3.9"  # optional since v1.27.0
...
```

To support this change to the specification I've made the following changes to the plugin:
* Allow `version` to be omitted from the compose file. When omitted, `version` is also omitted from the generated override compose file.
* Specifically check for version 1.* when verifying support for `cache_from`

Our main use case for this change is to improve how we manage our docker-compose files for use in integration tests. We typically layer multiple docker-compose files from different repos in the same docker-compose project to integration test a combination of services. When compose file versions are specified they must all be identical which makes it difficult to upgrade compose file versions and adds an unnecessary dependency across repos. Being able to omit them (and only specify a specific version when necessary) will make this easier for us to manage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following pipeline will run `test.sh` inside a `app` service container using
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.13.0:
+      - docker-compose#v4.0.0:
           run: app
 ```
 
@@ -28,7 +28,7 @@ through if you need:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.13.0:
+      - docker-compose#v4.0.0:
           run: app
           config: docker-compose.tests.yml
           env:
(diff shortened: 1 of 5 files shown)


jquick:
I ran into this same issue, this would be really nice to have!

pzeballos (approved):
great! This one will be a major release

PR #310, opened 2021-09-16
Add use-prior-image to completely skip repetitive builds

geoffharcourt:
This change adds a new plugin build option, `use-prior-image`. When set
to `true`, we'll check to see if a plugin with the expected image name
(the name specified by the cache-from parameter) exists on the
repository with `docker manifest inspect`. If it exists, we'll tag the
build metadata with the image name for the service and skip any further
pulling or building activity.

For builds where dependencies are built in a service, this can allow for
significant time shaved off of a blocking early build step.

### DIFF
--- a/plugin.yml
+++ b/plugin.yml
@@ -78,6 +78,8 @@ configuration:
       type: boolean
     mount-buildkite-agent:
       type: boolean
+    use-prior-image:
+      type: boolean
     entrypoint:
       type: string
     cli-version:
@@ -112,3 +114,4 @@ configuration:
     user: [ run ]
     propagate-uid-gid: [ run ]
     mount-buildkite-agent: [ run ]
+    use-prior-image: [ build ]
--- a/README.md
+++ b/README.md
@@ -336,6 +336,22 @@ steps:
 In the example above, the `myservice_intermediate:buildkite-build-${BUILDKITE_BUILD_NUMBER}` is one group named "intermediate", and `myservice:${BUILDKITE_BRANCH}` and `myservice:latest`
 are another (with a default name). The first successfully downloaded image in each group will be used as a cache.
 
+## Reusing previously built images
+
+If an build image has already been built previously, you can skip it if it
+already exists on the repository.
+
+```yaml
+steps:
+  - label: ":docker: Build an image"
+    plugins:
+      - docker-compose#v3.8.0:
+          build: app
+          image-repository: index.docker.io/myorg/myrepo
+          cache-from: app:index.docker.io/myorg/myrepo/myapp:123
+          use-prior-image: true
+```
+
 ## Configuration
 
 ### Main Commands
@@ -481,6 +497,11 @@ This option can also be configured on the agent machine using the environment va
 ### `cache-from` (optional, build only)
 
 A list of images to pull caches from in the format `service:index.docker.io/myorg/myrepo/myapp:tag` before building, ignoring any failures. If multiple images are listed for a service, the first one to successfully pull will be used. Requires docker-compose file version `3.2+`.
+
+### `use-prior-image` (optional, build only)
+
+When true, the build step will be skipped if the `cache-from` image exists on
+the remote repository. The run will use the existing image in later steps.
 
 ### `volumes` (optional, run only)
 
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -55,6 +55,16 @@ if [[ "$(plugin_read_config NO_CACHE "false")" == "false" ]] ; then
       fi
     fi
 
+    if [[ "$(plugin_read_config USE_PRIOR_IMAGE "false")" == "true" ]] ; then
+      if docker manifest inspect "$service_image" > /dev/null; then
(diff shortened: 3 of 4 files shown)


xzyfer:
This would be great. We currently hack around this by manually setting expect meta-data in a step prior to using the `docker-compose` plugin.

geoffharcourt:
We have now used this for almost two years across thousands of builds without any issues. It has dramatically speeded up some very repetitive steps in our build process.

lib/metadata.bash
```diff
@@ -20,6 +20,9 @@ function plugin_get_metadata() {
 function plugin_set_metadata() {
   local key="docker-compose-plugin-$1"
   local value="$2"
+
+  echo "buildkite-agent meta-data set $key $value";
+
```
toote, line 25: I assume these were debug lines and can be removed now, right?

plugin.yml
```diff
@@ -102,3 +104,4 @@ configuration:
     user: [ run ]
     propagate-uid-gid: [ run ]
     mount-buildkite-agent: [ run ]
+    use-prior-image: { build }
```
toote, line 107: I believe that, at least for consistency, these should be square brackets:
```suggestion
    use-prior-image: [ build ]
```

toote (changes requested):
In addition to my two (really minor comments) I would also suggest adding a test for when the `docker manifest inspect` fails.

Note that you will have to mergeback `master` on this PR for tests to succeed. My attempt would indicate it does so without conflicts but let me know if you want/need help or would rather I do that (assuming I have permissions to push on your branch)

plugin.yml
geoffharcourt: Thanks! I'm not sure if these changed, I've been rebasing this PR from time to time and just layering the changes onto what was upstream.

lib/metadata.bash
geoffharcourt: I think that's right.

plugin.yml
toote: From what I can see in the history of the repository, the square brackes have been used as early as 5 years ago :shrug:

geoffharcourt: Lo and behold, a second version of this on another branch did this the right way. No idea how this snuck in, but fixed now.

toote:
@geoffharcourt thanks a lot for the changes! 

I'm concerned about the `exit 0` in the plugin as it would mean that a step with multiple images to build will skip building if just the first one has a cache hit. I did a quick test changing that line to a `continue` but the test showed that the image tries to get built anyways. So I would suggest adding tests for that case and clearly defining the behaviour in those instances (you wouldn't want to skip builds for those that don't get a hit).

That said I'm curious about the exact use-case for this. Assuming this option does not exist or is turned off, regular behaviour would be that the `cache-from` stanzas would pull existing images to allow for layers to be re-used (so the build would be really quick) to finally have them tagged. My understanding is that you actually want to avoid the `pull` and `build` commands altogether and just get the images tagged for use in future steps. Is that correct?

toote:
@geoffharcourt in that case I would believe a more appropriate name would be `skip-build-if-cached` and would still need to factor in what to do if multiple `build` targets are specified and respecting other options like `build-alias`. I'll be more than happy to review your changes if you push them.

Having said that, a quick look I took around the code make it seem that there will be a lot of changes necessary. Which means that it may make more sense to have this kind of feature be its own separate thing. The quickest solution I can think of is a script that goes something like this:

```
# assume that there is a SERVICE_NAME
# assume that there are SERVICE_IMAGE_X to try to hit
SERVICE_IMAGES=()

for POSSIBLE_IMAGE in $(plugin_read_list SERVICE_IMAGE); do
  if docker manifest inspect "$POSSIBLE_IMAGE" > /dev/null; then
    echo "Found matching image $POSSIBLE_IMAGE, marking and skipping build"
    set_prebuilt_image "$SERVICE_NAME" "$POSSIBLE_IMAGE"
  else;
    SERVICE_IMAGES+=("$POSSIBLE_IMAGE")
  fi
done

# if we arrived here, there were no hits

# massage some variables here
SERVICE_IMAGES=$(IFS=',' echo "SERVICE_IMAGES[*]") # array to comma-delimited list

# upload a build step
buildkite-agent pipeline upload <<EOF
steps:
  - label: building $SERVICE_NAME
    plugins:
      - docker-compose#v3.12.0:
          build: $SERVICE_NAME
          cache-from: [ $SERVICE_IMAGES ]
EOF
```

Obviously, I'm oversimplifying things based on a lot of assumptions of your scenario (like you are only building a single service). Depending on the amount of options you are using on the build step the logic to re-create the build step can get tricky. But it may make for a great plugin on its own that would still give you all the benefits you are looking for and save you all the time you are investing in putting the square-like peg this plugin is through the round hole of your needs.

geoffharcourt:
Hi @toote I see what you're saying about the number of changes required here. At this point I think we've exposed that while this is really impactful for our use case, what I thought was a very generally applicable thing is in fact a very niche case. I appreciate your offer to review further changes, but I think we're just going to keep using similar logic as a private fork.

Issue #306, opened 2021-07-22
Support COMPOSE_FILE environment variable


evandam:
Hi folks,

It would be a nice improvement to be able to set [`COMPOSE_FILE`](https://docs.docker.com/compose/reference/envvars/#compose_file) at the top of a pipeline rather than remembering to set it in each step that uses this plugin, but it seems this plugin overrides to `docker-compose.yml`.

For example:

```yml
env:
  COMPOSE_FILE: docker-compose-buildkite.yml
  DOCKER_IMAGE: quay.io/wonolo/platform-task-scheduler

steps:
- label: ":docker: build dev image"
  key: build-dev
  plugins:
  - docker-compose#v3.8.0:
      # config: docker-compose-buildkite.yml
      build: app
      image-repository: $DOCKER_IMAGE
      image-name: dev-${GIT_SHA}

- label: ":docker: push dev tags"
  key: push-dev-tags
  plugins:
  - docker-compose#v3.8.0:
      # config: docker-compose-buildkite.yml
      push:
      - app:${DOCKER_IMAGE}:dev-latest
  depends_on:
  - build-dev

- label: ":rspec: RSpec"
  key: rspec
  command: bundle exec rspec --format progress
  plugins:
  - docker-compose#v3.8.0:
      # config: docker-compose-buildkite.yml
      run: app
      entrypoint: ""
  depends_on:
  - build-dev
```

PR #311, opened 2021-09-28
read $COMPOSE_FILE for default config file if present

evandam:
Fixes https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/306

Thoughts? @toolmantim?


jorge-wonolo:
Yes please this would be really useful!

PR #325, opened 2022-04-15
Allow build and run

jeremybumsted:
This will take the changes made in #244 and add some tests to validate the functionality added by allowing both `build` and `run`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following pipeline will run `test.sh` inside a `app` service container using
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.11.3:
+      - docker-compose#v3.12.0:
           run: app
 ```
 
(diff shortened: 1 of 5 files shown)


toote:
I have updated this PR with the latest changes and updated the plugin.yml as well as part of the readme... and while doing so I found that the [run step already has some logic for building images](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/a3b35d1fe23265d9f70e7ccae408052aacad45e8/commands/run.sh#L236) so that behaviours really needs to be tested before we can merge this.

toote:
I have done a mergeback from master, added a lot of tests and re-worked the documentation a bit. Finally, prepared the examples for the next feature release (which should be -at least- this one)

pzeballos (approved):

Issue #331, opened 2022-08-16
add mount-checkout option from docker-buildkite-plugin


TSMMark:
Hi, would it be possible to add the `mount-checkout` or `copy-checkout` option from `docker-buildkite-plugin` implemented here:

https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/182

Or maybe this is already possible to do and undocumented? I think this would save a ton of time on certain builds

TSMMark:
Basically I want to cp files from the host into the container, and I want to avoid docker volumes.

Any way to use docker compose cp command directly using `docker-compose-buildkite-plugin`? https://docs.docker.com/engine/reference/commandline/compose_cp/

toote:
This is now part of the latest release v4.4.0 :)

PR #332, opened 2022-08-25
POC Run docker compose cp

TSMMark:
Hi, this does a `docker cp` command to get project files into a docker compose container.

# Why?

Docker volumes have been a huge perf issue for us so I want to avoid volumes.

And doing a `COPY .` directive in the dockerfile adds a ton of extra size to the docker image, and seems incredibly redundant since the files are **already on disk from the git checkout build step.**

# How?

Simply `cp` the files into the container after the container boots

In my project's pipeline.yml, wait until the file exists before running the `command` script

```
command: ['/bin/sh', '-c', 'until [ -f ./scripts/ci_smartlink_cypress_test.sh ]; do sleep 1; done; ./scripts/ci_smartlink_cypress_test.sh']
```

---

This PR needs some cleanup and adding some config options to specify which files to copy, and removing some of the hardcoded values.

If maintainers think this is something that could be merged I can work on cleaning it up. Let me know

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -253,8 +253,13 @@ if [[ "$(plugin_read_config DEPENDENCIES "true")" == "true" ]] ; then
 fi
 
 # TODO: Customize which files are copied to the container
-echo "--- Copying . into ${run_service}:/vydia/"
-docker cp . "${run_service}:/vydia/"
+echo "--- :docker: Copying . into ${run_service}:/vydia/"
+echo "docker ps"
+docker ps
+container_id="$(run_docker_compose ps -q "${run_service}")"
+echo "docker container_id: ${container_id}"
+echo "docker cp"
+docker cp . "${container_id}:/vydia/"
 echo "Done cp"
 
 shell=()
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -254,7 +254,7 @@ fi
 
 # TODO: Customize which files are copied to the container
 echo "--- Copying . into ${run_service}:/vydia/"
-run_docker_compose cp . "${run_service}:/vydia/"
+docker cp . "${run_service}:/vydia/"
 echo "Done cp"
 
 shell=()
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -252,6 +252,11 @@ if [[ "$(plugin_read_config DEPENDENCIES "true")" == "true" ]] ; then
   echo
 fi
 
+# TODO: Customize which files are copied to the container
+echo "--- Copying . into ${run_service}:/vydia/"
+run_docker_compose cp . "${run_service}:/vydia/"
+echo "Done cp"
+
 shell=()
 shell_disabled=1
 result=()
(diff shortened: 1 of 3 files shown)


toote:
Hi there, thanks for the PR! interesting idea but I believe that a volume mount of the checked out directory (as you mentioned in #331) would be simpler to implement and test. Or is there any reason in particular that the `cp` is preferred?

TSMMark:
Yes, I tried that originally and I found that using a docker volume severely crippled the disk IO performance

toote:
@TSMMark makes sense... feel free to continue working on this and we will be more than happy to provide review and guidance. In the meantime I would suggest you move this PR to be a draft until you believe it is ready. What do you think?

pda:
> Yes, I tried that originally and I found that using a docker volume severely crippled the disk IO performance

I'm curious what might cause that reduced filesystem performance.

My (possibly outdated or oversimplified) understanding of a docker volume is that it's a bind mount and the kernel level, which shouldn't introduce any performance implications vs accessing that directory from outside the container, nor differ from accessing files within the container.

pzeballos:
Hey @TSMMark! Can we close this PR or are you still working on it? you can always open a new one when there's more progress

TSMMark:
We've been using this fork since it was created and it's working great. I wasn't given any clear indication that this would ever be accepted so I have no intention on making this parameterized & configurable if it won't be accepted anyway. If that's not the case, let me know if it might be accepted and how you'd like it to be configured from pipeline.

toote:
Hi @TSMMark!

As far as I understand, the added script is ran in the background and waits for the main container (`${run_container}`) to be up and running before doing a `docker cp` to it. The container itself is running a busy-loop so that it waits for the files to be available before it actually does anything (thus avoiding any race conditions or failures due to missing files).

While I think that there may be others with your same issue and agree that it would be great for the plugin to offer a solution I can not think of a way to make your code work for the general case. There are a lot of moving parts that have to be taken into account. Off the top of my head, the following changes will be necessary to the code:

* functionality should be encapsulated in a plugin option (or maybe more than one)
* the script should only be called if the option is enabled (and maybe within the same subshell as the main container is ran)
* incompatible options (like the one that mounts the workspace) should raise errors
* checks must be added to prevent some weird scenarios (like copying the files into another docker volume)
* the command ran in the container must be adapted automatically to wait for the files to exist (probably adding a sentinel file and ensure that it is copied last) whether it was a command set in the buildkite step or the plugin's own `command` option
* tests must be added (which may be a lot tricker than expected due to the processes ran in the background)

I know that this is a lot of work for something that is already working for you so we would totally understand if you are unable (or unwilling) to devote the time necessary - specially if you take into account that there will probably be some back-and-forth during code review. Which is why we would also be OK if you decide not to and just close the PR (or ask us to do so); the fact that it existed and all these discussions should be enough for others in your same situation to tailor the plugin to their needs. And, who knows, maybe someone else will create a future PR!

PR #334, opened 2022-09-16
Add support docker compose v2

donbobka:
The new version (v2) of docker-compose gives a lot of new features, especially recently added `secrets` and `platforms` options to docker-compose file. It will be nice to be able to use it with docker-compose-buildkite-plugin.

docker-compose v2 and docker-compose v1 CLI are not fully compatible in some places + coexistable and that's why the proposal is to keep it as an opt-in feature: by default plugin will use `docker-compose` with a possibility to switch to `docker compose` executable

```
steps:
  - command: generate-dist.sh
    plugins:
      - docker-compose#v3.10.0:
          build: app
          cli-version: 2
          image-repository: index.docker.io/myorg/myrepo
          args:
            - MY_CUSTOM_ARG=panda
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,10 @@ Select when to upload container logs.
 
 The default is `on-error`.
 
+### `cli-version` (optional)
+
+If set to `1`, plugin will use `docker-compose` executable, else `docker compose` will be used.
+
 ## Developing
 
 To run the tests:
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following pipeline will run `test.sh` inside a `app` service container using
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           run: app
 ```
 
@@ -28,7 +28,7 @@ through if you need:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           run: app
           config: docker-compose.tests.yml
           env:
@@ -41,7 +41,7 @@ or multiple config files:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           run: app
           config:
             - docker-compose.yml
@@ -56,7 +56,7 @@ env:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           run: app
 ```
 
@@ -67,7 +67,7 @@ steps:
   - plugins:
       - docker-login#v2.0.1:
           username: xyz
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           build: app
           image-repository: index.docker.io/myorg/myrepo
   - wait
@@ -75,7 +75,7 @@ steps:
     plugins:
       - docker-login#v2.0.1:
           username: xyz
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           run: app
 ```
 
@@ -84,7 +84,7 @@ If you want to control how your command is passed to docker-compose, you can use
 ```yml
 steps:
   - plugins:
-      - docker-compose#v3.11.1:
+      - docker-compose#v3.11.2:
           run: app
           command: ["custom", "command", "values"]
 ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -479,6 +479,10 @@ Select when to upload container logs.
 
 The default is `on-error`.
 
+### `cli-version` (optional)
+
+If set to `1`, plugin will use `docker-compose` executable, else `docker compose` will be used.
```
toote, line 484: I believe that his wording makes it sound like version 2 is the default (I actually had to read the code to confirm it was not the case). I may suggest to word it differently:

```suggestion
If set to `2`, plugin will use `docker compose` to execute commands; otherwise it will default to version `1` using `docker-compose` instead.
```

Note that this wording also means that you will have to make a minor change to the logic of the code (see my other comment)

lib/shared.bash
```diff
@@ -173,7 +173,12 @@ function build_image_override_file_with_version() {
 
 # Runs the docker-compose command, scoped to the project, with the given arguments
 function run_docker_compose() {
-  local command=(docker-compose)
+  local command
+  if [[ "$(plugin_read_config CLI_VERSION "1")" == "1" ]] ; then
+    command=(docker-compose)
+  else
+    command=(docker compose)
+  fi
```
toote, line 181: I believe it would be simpler (and more in line with my other comment to do this instead):

```suggestion
  local command=(docker-compose)
  if [[ "$(plugin_read_config CLI_VERSION "1")" == "2" ]] ; then
    command=(docker compose)
  fi
```

toote (changes requested):
Very nice and simple PR... you are missing the changes to the `plugin.yml` file though to ensure that the option is validated correctly (make sure to list valid values for the opion). Just for completeness sake I would also suggest adding a test when the version is set to 1 explicitly

toote:
@donbobka please note that you will need a mergeback from `master` to fix the tests (and correct a race condition we found in the pipeline)... we are more than happy to do that for you - it was a big change we recently introduced so no point in putting that on you ;) - before merging if you believe that your code is now ready

donbobka:
@toote All done! Thank you for your review.

P.S.: I still testing it on a real project. So far everything works (build+push), but I found at least one example when the logic can differs between versions. One case that I found: in v2 we can skip pull image when cache-from is used.

toote:
> One case that I found: in v2 we can skip pull image when cache-from is used.

The fact that it _can_ be skipped doesn't mean that it should if it means that the change in the PR is smaller. As long as nothing breaks on existing functionality, there is no need to optimize every scenario.

But you do mention something that I didn't realize, it may be a great idea to copy the `run`, `build` and `push` tests to new test files and have a copy of all tests when the version is changed. If you run the tests with `docker run --rm -ti -v $PWD:/plugin -e DOCKER_STUB_DEBUG=3 bats tests/YOUR_FILE` you should be able to see exactly what is being run on `YOUR_FILE` tests when the `docker` stub command is called

donbobka:
@toote I copied all run, build, push tests into v2 folder and updated them to work with docker compose v2. I have to add `-r` option [here](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/334/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R25) to include v2 subdirectory.

PS: Maybe adding `-r` to [buildkite-plugin-tester](https://github.com/buildkite-plugins/buildkite-plugin-tester/blob/main/Dockerfile#L44) is better idea

toote:
@donbobka adding `-r` one has to be careful with the files added to the `tests` subfolders. There are two other alternatives: have them all be files in the main folder with a `v2` prefix or suffix, or maybe even have v2 tests run in a completely separate step (so you run `bats /tests/v2` without the need for a `-r`.

That said it is indeed an good suggestion, though probably a functionality for the plugin tester through an environment variable instead

donbobka:
I kept one step, but now explicitly specified `tests/v2`

toote (approved):
Awesome job :)

TSMMark:
This is great, but I can't figure out — how does one configure docker buildkit secrets using this?

e.g. `--secret` flag described here https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information

toote:
@TSMMark as far as I understand there is no support for the `--secret` flag in this plugin yet

TSMMark:
@toote Thanks for clarifying. 

I do think using buildkit secrets would be a big win for the security of containers running on buildkite. AFAICT currently the quickest way to provide secrets to a docker build using this plugin is with build args + env vars — however, those approaches expose the secrets in plain text in the built docker image and is considered bad practice for obvious security reasons

PR #335, opened 2022-09-17
Update plugin tester and tests

toote:
Updates the plugin tester and fixes tests that broke (originally a lot but it was mostly due to [the bug in the tester release](https://github.com/buildkite-plugins/bats-mock/pull/8)).

If there is a new re-lease of the plugin tester soon, we should wait for it before merging/releasing this one

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 version: '2'
 services:
   tests:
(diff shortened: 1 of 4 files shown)


pzeballos (approved):
LGTM

PR #337, opened 2022-09-23
Warnings with step-level array commands

### LINKED ISSUE #312: `exit 1` Even though commands passed
The plugin is reliably exiting 1, and hence failing my runs, with rather long runtimes. I have a 4 hour test which triggers this every time.

My pipeline definition looks like:
```
agents:
  queue: 'default'
  docker: '*'

steps:
  - label: 'nightly-test'
    timeout_in_minutes: 360
    concurrency: 1 # Only allow one nightly build at a time
    concurrency_group: 'node-sdk/nightly'
    commands:
      - '.buildkite/docker/build.sh'
      - 'npm run ci-nightly'
      - 'echo success'
    artifact_paths:
      - 'docker-compose-logs/*'
    plugins:
      - docker-compose#v3.8.0:
          run: test-container
          upload-container-logs: always
          config:
            - .buildkite/docker/docker-compose.yaml
            - .buildkite/docker/docker-compose-ci.yaml
```

I know the commands work fine because I see the `success` line get echoed. The build output ends with:
```
success
🚨 Error: The command exited with status 1
user command error: The plugin docker-compose command hook exited with status 1
```

Whereas builds that have failed for legitimate reasons look like:
```
⚠️ Failed to run command, exited with 2, run params:
run --name buildkitec14d1406904d4db98fa70cefd10a4e75_test-container_build_250 --rm test-container /bin/sh -e -c .buildkite/docker/build.sh
npm run ci-load
🚨 Error: The command exited with status 2
user command error: The plugin docker-compose command hook exited with status 2
```

The "failed to run command" part being missing says to me the plugin isn't mistakenly thinking my command failed. It thinks it passed, and then exits 1 anyway.

I haven't dug into plugin code yet but this is preventing me from using the plugin for these longer running tests.

toote:
Based on previous discussions (#59, #186 and #312), this plugin does not support step-level commands defined as an array. We just receive a multi-line string in the `BUILDKITE_COMMAND` variable but there is no way to identify it is because a single command with newlines (like [we test in the pipeline](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/2da89a5fcbcb4b8c534d40b978b77d50acdee2c8/.buildkite/pipeline.yml#L44-L47)) or an actual array to be able to manipulate the string and do what we are supposed to do (like what the agent does: creates a script enclosing the whole command and runs that script instead).

So the only possible alternative is to add a warning to the documentation and output for people to be aware of the existing limitation :shrug: 

Closes #312

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ steps:
           run: app
 ```
 
+:warning: Warning: you should not use this plugin with an array of commands at the step level. Execute a script in your repository, a single command separated by `;` or the plugin's [`command` option](#command-optional-run-only-array) instead.
+
 You can also specify a custom Docker Compose config file and what environment to pass
 through if you need:
 
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -339,6 +339,12 @@ if [[ ${#shell[@]} -gt 0 ]] ; then
 fi
 
 if [[ -n "${BUILDKITE_COMMAND}" ]] ; then
+  if [[ $(echo "$BUILDKITE_COMMAND" | wc -l) -gt 1 ]]; then
+    # An array of commands in the step will be a single string with multiple lines
+    # This breaks a lot of things here so we will print a warning for user to be aware
+    echo "⚠️  Warning: The command received has multiple lines."
+    echo "⚠️           The Docker Compose Plugin does not correctly support step-level array commands."
+  fi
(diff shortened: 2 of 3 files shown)


pzeballos (approved):
Love the warning :P

PR #341, opened 2022-10-06
Skip pull in run step

### LINKED ISSUE #232: skip-checkout doesn't seem work as expected

### LINKED ISSUE #292: Option to skip pulling images?
Hi folks,

In light of DockerHub's rate limiting, is it possible to skip attempting to pull images every time before running a step? (https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/commands/run.sh#L73)

As far as I know, docker-compose will pull the image down if it needs to, or otherwise just use the local image.

It would be great to have something like `force_pull: false`.

toote:
Implements the `skip-pull` option

Closes #292 

Also adds a minor clarification to the `skip-checkout` option.

Closes #232


pzeballos (approved):
🙌🏻

PR #344, opened 2022-10-10
Propagate environment

### LINKED ISSUE #61: Automatically propagate environment variables into the container
When setting pipeline environment variables it commonly catches me and other people in our company out that these variables are not shared with the container. We can specify them manually in the docker-compose file, but it would be nice if this was done automatically.

toote:
Port the `propagate-environment` option from the docker plugin

Closes #61


pzeballos (approved):
🎉

PR #346, opened 2022-10-12
Add `target` option

### LINKED ISSUE #231: Add support for overriding `target`
Target being: https://docs.docker.com/compose/compose-file/#target

Allowing you to override this using the plugin would make it possible for you to build multiple targets of a service, push each target to a repository, and then pull them down in a `cache-from` in order to use caching for multi-stage builds. This is still doable today, but you have to have multiple docker-compose.yml files for each target. Adding `target` to the plugin would make it much easier.

toote:
As described.

While I was at it, I sorted the options in the `plugin.yml` file and corrected all shellcheck warnings in tests.

Closes #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,7 +464,7 @@ Whether to skip the repository checkout phase. This is useful for steps that use
 
 ### `skip-pull` (optional, run only)
 
-Completely avoid running any `pull` command. Images being used will need to be present in the machine from before or have been built in the same step. Could be useful to avoid hitting rate limits when you can be sure the operation is unnecessary. Note that it is possible other commands run in the plugin's lifecycle will trigger a pull of necessary images. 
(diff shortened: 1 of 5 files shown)


pzeballos (approved):

viraptor:
Does this apply to a specific version of docker-compose? I'm trying to track down how the option works, but `docker compose build` does not seem to accept `--target` at all.
https://docs.docker.com/engine/reference/commandline/compose_build/ does not list the `--target` option.

agibralter:
This is a random idea, but regarding: `Note that there is a single build command run for all services so the target value will apply to all of them.`

What if instead of a single target configuration for the whole step, you could specify something like this:

```yaml
- key: "test"
  plugins:
    docker-compose#v5.3.0:
      build:
        - service-name:target-stage
```

Or some other delimiter that would make sense:
- `service-name@target-stage`
- `service-name,target-stage`
- `service-name;target-stage`
- `service-name|target-stage`

PR #347, opened 2022-10-15
BuildKit & `ssh` support

toote:
Adds two options:
* `buildkit` to automatically setup the environment variables to enable BuildKit support if it is available
* `ssh`, to enable the ability to selectively share the ssh agent environment in a build step (that only works if BuildKit is enabled)


pzeballos (approved):
💚

PR #348, opened 2022-10-17
Add mount-checkout support

francoiscampbell:
Addresses https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/331 by copying the behaviour of the Docker plugin to automatically mount the checkout into the main container.

If users wish to copy files instead of using a volume, they can instead use https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/332 once that's merged.

~Question for maintainers: this enables the `mount-checkout` option by default to keep the behaviour the same as the Docker plugin. This also requires that when the option is enabled, either manually or by the default, that `--workdir` also be passed. I feel like this is a breaking change, so should this PR result in a new major version or should I make the default `false` and break parity with the Docker plugin? (I suppose it could still be made `true` by default in a future major version)~ Edit: made it `false` by default.

This is working as-is in the default case, but I'll keep this as draft while I add tests and perform manual testing in our pipelines, and I'm open to reviews and comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,11 +494,11 @@ Whether to automatically mount the `buildkite-agent` binary and associated envir
 
 Default: `false`
 
-### `mount-checkout` (optional, boolean)
+### `mount-checkout` (optional, run-only, boolean)
 
 Whether to automatically mount the current working directory which contains your checked out codebase. Mounts onto `/workdir`, unless `workdir` is set, in which case that will be used.
 
-Default: `true`
+Default: `false`
 
 ### `pull-retries` (optional)
 
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ Completely avoid running any `pull` command. Images being used will need to be p
 
 ### `workdir` (optional, run only)
 
-Specify the container working directory via `docker-compose run --workdir`. The default is `/workdir`. This path is also used by `mount-checkout` to determine where to mount the checkout in the container.
+Specify the container working directory via `docker-compose run --workdir`. This option is also used by [`mount-checkout`](#mount-checkout-optional-run-only-boolean) to determine where to mount the checkout in the container.
 
 Example: `/app`
 
--- a/README.md
+++ b/README.md
@@ -468,7 +468,9 @@ Completely avoid running any `pull` command. Images being used will need to be p
 
 ### `workdir` (optional, run only)
 
-Specify the container working directory via `docker-compose run --workdir`.
+Specify the container working directory via `docker-compose run --workdir`. The default is `/workdir`. This path is also used by `mount-checkout` to determine where to mount the checkout in the container.
+
+Example: `/app`
 
 ### `user` (optional, run only)
 
@@ -491,6 +493,12 @@ Default: `false`
 Whether to automatically mount the `buildkite-agent` binary and associated environment variables from the host agent machine into the container.
 
 Default: `false`
+
+### `mount-checkout` (optional, boolean)
+
+Whether to automatically mount the current working directory which contains your checked out codebase. Mounts onto `/workdir`, unless `workdir` is set, in which case that will be used.
+
+Default: `true`
 
 ### `pull-retries` (optional)
 
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -9,7 +9,7 @@ container_name="$(docker_compose_project_name)_${run_service}_build_${BUILDKITE_
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 pull_retries="$(plugin_read_config PULL_RETRIES "0")"
 mount_ssh_agent=''
-mount_checkout="$(plugin_read_config MOUNT_CHECKOUT "true")"
+mount_checkout="$(plugin_read_config MOUNT_CHECKOUT "false")"
 
 expand_headers_on_error() {
   echo "^^^ +++"
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -9,6 +9,7 @@ container_name="$(docker_compose_project_name)_${run_service}_build_${BUILDKITE_
 override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
 pull_retries="$(plugin_read_config PULL_RETRIES "0")"
 mount_ssh_agent=''
+mount_checkout="$(plugin_read_config MOUNT_CHECKOUT "true")"
 
 expand_headers_on_error() {
   echo "^^^ +++"
@@ -115,10 +116,16 @@ if [[ -n "${BUILDKITE_REPO_MIRROR:-}" ]]; then
 fi
 
 tty_default='true'
+workdir_default="/workdir"
+pwd_default="$PWD"
 
 # Set operating system specific defaults
 if is_windows ; then
   tty_default='false'
+  workdir_default="C:\\workdir"
+  # escaping /C is a necessary workaround for an issue with Git for Windows 2.24.1.2
+  # https://github.com/git-for-windows/git/issues/2442
+  pwd_default="$(cmd.exe //C "echo %CD%")"
 fi
 
 # Optionally disable allocating a TTY
@@ -131,8 +138,19 @@ if [[ "$(plugin_read_config DEPENDENCIES "true")" == "false" ]] ; then
   run_params+=(--no-deps)
 fi
 
-if [[ -n "$(plugin_read_config WORKDIR)" ]] ; then
-  run_params+=("--workdir=$(plugin_read_config WORKDIR)")
+workdir=''
+
+if [[ -n "$(plugin_read_config WORKDIR)" ]] || [[ "${mount_checkout}" == "true" ]]; then
+  workdir="$(plugin_read_config WORKDIR "$workdir_default")"
+fi
+
+if [[ -n "${workdir}" ]] ; then
+  run_params+=("--workdir=${workdir}")
+fi
+
+# By default, mount $PWD onto $WORKDIR
+if [[ "${mount_checkout}" == "true" ]] ; then
+  run_params+=("-v" "${pwd_default}:${workdir}")
 fi
 
 # Can't set both user and propagate-uid-gid
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -1159,3 +1159,84 @@ export BUILDKITE_JOB_ID=1111
   unstub docker-compose
   unstub buildkite-agent
 }
+
+@test "mount-checkout is not enabled by default" {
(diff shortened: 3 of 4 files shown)


commands/run.sh
```diff
@@ -131,8 +136,19 @@ if [[ "$(plugin_read_config DEPENDENCIES "true")" == "false" ]] ; then
...
+  workdir="$(plugin_read_config WORKDIR "$workdir_default")"
+fi
+
+if [[ -n "${workdir}" ]] ; then
+  run_params+=("--workdir=${workdir}")
+fi
+
+# By default, mount $PWD onto $WORKDIR
+if [[ "${mount_checkout}" == "true" ]] ; then
+  run_params+=("-v" "${pwd_default}:${workdir}")
```
francoiscampbell, line 151: basically lifted from https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/391ee61fb87b1281d85e97263fd4a46181265687/hooks/command#L127-L136 and https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/391ee61fb87b1281d85e97263fd4a46181265687/hooks/command#L193-L196 and adapted to this plugin

commands/run.sh
```diff
@@ -115,10 +116,14 @@ if [[ -n "${BUILDKITE_REPO_MIRROR:-}" ]]; then
 fi
 
 tty_default='true'
+workdir_default="/workdir"
+pwd_default="$PWD"
```
francoiscampbell, line 120: https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/391ee61fb87b1281d85e97263fd4a46181265687/hooks/command#L81-L82

commands/run.sh
```diff
@@ -115,10 +116,16 @@ if [[ -n "${BUILDKITE_REPO_MIRROR:-}" ]]; then
...
 
 # Set operating system specific defaults
 if is_windows ; then
   tty_default='false'
+  workdir_default="C:\\workdir"
+  # escaping /C is a necessary workaround for an issue with Git for Windows 2.24.1.2
+  # https://github.com/git-for-windows/git/issues/2442
+  pwd_default="$(cmd.exe //C "echo %CD%")"
```
francoiscampbell, line 128: https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/391ee61fb87b1281d85e97263fd4a46181265687/hooks/command#L89-L92

README.md
```diff
@@ -492,6 +494,12 @@ Whether to automatically mount the `buildkite-agent` binary and associated envir
 
 Default: `false`
 
+### `mount-checkout` (optional, boolean)
+
+Whether to automatically mount the current working directory which contains your checked out codebase. Mounts onto `/workdir`, unless `workdir` is set, in which case that will be used.
+
+Default: `true`
```
toote, line 501: I would definitely set it to `false` as default. I don't think there is a need to make this change backwards-incompatible when it would be the same amount of effort with no risk to make it so.

README.md
```diff
@@ -492,6 +494,12 @@ Whether to automatically mount the `buildkite-agent` binary and associated envir
 
 Default: `false`
 
+### `mount-checkout` (optional, boolean)
```
toote, line 497: it is important to note that this is only for the `run` stages:
```suggestion
### `mount-checkout` (optional, run only, boolean)
```

toote (changes requested):
Great work and thanks for the PR!

I do believe that making it backwards-incompatible makes no sense so I would suggest to switching it to false. That way you will also avoid having to modify existing tests :P I would also suggest you add a dependency on the `run` property for the option in the `plugin.yml` file. 

Other than that, remove it from draft when tests are done and I'll review it asap :)

francoiscampbell:
Thanks for the review @toote, added some tests

tests/run.bats
```diff
@@ -1159,3 +1159,84 @@ export BUILDKITE_JOB_ID=1111
...
+    "meta-data exists docker-compose-plugin-built-image-tag-myservice : exit 0" \
+    "meta-data get docker-compose-plugin-built-image-tag-myservice : echo myimage"
+
+  run "$PWD"/hooks/command
+
+  assert_success
+  assert_output --partial "ran myservice without mount-checkout"
+  unstub docker-compose
+  unstub buildkite-agent
+}
```
toote, line 1187: While I value completeness, this test is not really needed as every other test already tests this :P

On the other hand, if you added a `export BUILDKITE_PLUGIN_DOCKER_COMPOSE_MOUNT_CHECKOUT=false` it would indeed be a necessary test.

toote (changes requested):
One minor change on the tests... but the rest of the whole thing is great!

tests/run.bats
francoiscampbell: Thanks for the feedback, updated

README.md
```diff
@@ -468,7 +468,9 @@ Completely avoid running any `pull` command. Images being used will need to be p
 
 ### `workdir` (optional, run only)
 
-Specify the container working directory via `docker-compose run --workdir`.
+Specify the container working directory via `docker-compose run --workdir`. The default is `/workdir`. This path is also used by `mount-checkout` to determine where to mount the checkout in the container.
```
toote, line 471: Was doing a final review and found this little change to make (as the default only applies when `mount-checkout` is used, so the documentation in that option should be enough:
```suggestion
Specify the container working directory via `docker-compose run --workdir`. This option is also used by [`mount-checkout`](#mount-checkout-optional-run-only-boolean) to determine where to mount the checkout in the container.
```

toote (approved):
Approved because I don't think its fair not to when the only comment left is a minor documentation one and it would be like the fifth back-and-forth :P

Really appreciate the contribution and your patience with all my nagging :+1:

TSMMark:
Thanks for adding this, it looks great. I'm trying it out now. 

Running into an issue where the mounted dir is removing existing files in the directory.

For example, if my Dockerfile contains files in `/app`:

```
COPY .yarnrc package.json yarn.lock /app/
WORKDIR /app
RUN yarn install --frozen-lockfile && yarn cache clean
```

and plugin options specify `/app`
 
```
      - docker-compose#v4.4.0:
          mount-checkout: true
          workdir: /app
```

The mounted files seem to be replacing the directory contents and removing the `node_modules` directory that was added during `yarn install` in docker build.

Is this expected? Would a rename and then `mv` strategy work? 

E.g.

1. mount at `/app_tmp`
2. docker exec `mv /app_tmp /app`

I think that might work, and the existing `/app` contents would be retained.

toote:
@TSMMark you bring up a valid point and I created a PR to change the way `mount-checkout` works so that you can setup `workdir` and `mount-checkout` and they are both mounted at different places (see #354)

That said, docker does not allow for "merging" of folders being mounted either between each other nor with content that already existed in the image. If you need both folders, you will have to mount them in separate paths and then work with them as you want/need.

TSMMark:
got it merging the folders using rsync in a shell script in the entrypoint cmd, like:

```
SOURCE_DIR=/vydia_tmp_mount
TARGET_DIR=/vydia
rsync -ah "$SOURCE_DIR/" "$TARGET_DIR/"
```

It seems to work, but it's inconsistent and seems to error occasionally, sometimes timing out, sometimes hanging forever until the step times out and receives cancellation signal.

"Read timed out": 

```
[2022-10-24T19:16:35Z] Creating buildkite01840b6ab59f4e28b1959313542659bb_ci_run ...
  | [2022-10-24T19:17:35Z]
  | [2022-10-24T19:17:35Z] ERROR: for buildkite01840b6ab59f4e28b1959313542659bb_ci_run  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
  | [2022-10-24T19:17:35Z]
  | [2022-10-24T19:17:35Z] ERROR: for ci  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
  | [2022-10-24T19:17:35Z] ERROR: An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
  | [2022-10-24T19:17:35Z] If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).
```

"hang forever":

```
[2022-10-24T19:20:45Z] $ docker-compose -f docker-compose.ci.yml -p buildkite01840b70837b4e19a00a77f109318094 -f docker-compose.buildkite-23033-override.yml run --name buildkite01840b70837b4e19a00a77f109318094_ci_build_23033 -e CODECOV_TOKEN -e PRIVATE_SSH_KEY -e COVERAGE=true --no-deps --workdir=/vydia -v /var/lib/buildkite-agent/builds/buildkite-spot-i-0426dafd8fd1a8b10-1/vydia/web:/vydia_tmp_mount --rm ci /bin/sh -c 'until [ -d codemods ]; do sleep 1; done; cd codemods && yarn install --frozen-lockfile && yarn test'
[2022-10-24T19:20:46Z] Creating network "buildkite01840b70837b4e19a00a77f109318094_default" with the default driver
[2022-10-24T19:20:46Z] Creating buildkite01840b70837b4e19a00a77f109318094_ci_run ...
```

Not sure what would cause this

toote:
@TSMMark the error you mention appears to be a network timeout in your installation that doesn't appear to be related with this feature in particular but with your environment (the docker daemon stopped responding for some reason).

TSMMark:
Sure but it happened repeatedly across multiple agents, and only on builds using the mount-checkout feature so there is almost definitely a correlation

elikor-figma:
@TSMMark Were you ever able to find a workaround for this issue? I am also encountering the "Read timed out" error when using mount-checkout.

TSMMark:
@elikor-figma I've got this PR open as a POC https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/332

I've been using this fork for 7 months and it's solved all my issues, like the Read timed out error with volumes, and also avoids having to rebuild + push the docker image every time the source code changes (basically every commit)

I didn't spend the time parameterizing it, so it has the `vydia` value hardcoded in the PR, but basically you could fork that and change `vydia` to the name of the directory you want to copy when the container starts up.

Then reference your fork commit in pipeline.yml, like:

```
    plugins:
      - elikor-figma/docker-compose#your-commit-hash-abcdefg12345
```

You may have to update your commends in pipeline.yml to wait for the file to exist before executing the script

```
command: ['/bin/sh', '-c', 'until [ -f ./scripts/ci_db_annotate.sh ]; do sleep 1; done; ./scripts/ci_db_annotate.sh']
```

---

If this approach works for you , feel free to comment on that PR showing support. It so far has no traction

TSMMark:
@elikor-figma curious if you had any success using the `cp` approach?

PR #350, opened 2022-10-17
Default tag in `cache-from`

### LINKED ISSUE #349: cache-from option requires tag
With the changes in v4.1.1, the plugin is now erroring if a tag is not defined in the `cache-from` configuration option.

I believe before this change it had the same functionality as `docker pull`, where if a tag was not specified it would fallback on latest.

toote:
Allows tag-less `cache-from` stanzas

Closes #349


pzeballos (approved):
good catch!

PR #355, opened 2022-10-28
Wait for cluster to be healthy

usmanismail:
Currently the compose plugin is not waiting for the services to be healthy. 
This is causing problems in our CI where some containers need warmup time

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -268,9 +268,9 @@ if [[ "$(plugin_read_config DEPENDENCIES "true")" == "true" ]] ; then
   # Start up service dependencies in a different header to keep the main run with less noise
   echo "~~~ :docker: Starting dependencies"
   if [[ ${#up_params[@]} -gt 0 ]] ; then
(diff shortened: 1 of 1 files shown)


toote:
it is indeed an interesting use case, unfortunately, the change broke existing tests with dependencies.

How about adding it behind an option that defaults to `false`? This not only would avoid a major release, but it would also mean that existing tests wouldn't break and you would just need to add a couple new ones only. What do you think?

toote:
@usmanismail let us know if you are still interested in working on this functionality. If you are not but still want this functionality, either [allow us permissions to push on this PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) or let us know and close this PR so that we can open a new one :)

usmanismail:
Hi Sorry, 
I found a cleaner solution for my issue i.e. Using a `condition: service_healthy` inside the docker compose file. 
Feel free to close and create a new PR or repurpose this one as needed.

toote:
Thanks a lot for your contribution anyhow, I believe that this may be a good feature anyways for those that can't modify their docker compose file so we'll add it as a feature :)

PR #362, opened 2022-11-16
Add support for exposing service ports on run

trvrnrth:
This stems from a need for a step run in a container to have a port exposed that can be connected to from an external testing service whilst it is running.

By default `docker compose run` does not create the service ports so this adds a configuration to optionally to add the `--service-ports` [option](https://docs.docker.com/engine/reference/commandline/compose_run/#options) which enables the service ports and maps them to the host.

The default run configuration is not affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -606,6 +606,12 @@ The default is `true`.
 
 Sets the `--entrypoint` argument when running `docker-compose`.
 
+### `service-ports` (optional, run only)
+
+If set to true, docker compose will run with the service ports enabled and mapped to the host. Equivalent to `--service-ports` in docker-compose.
+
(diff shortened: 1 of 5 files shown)


toote (changes requested):
Thanks a lot for the PR @trvrnrth, it is almost perfect.

A minor change I would suggest would be to add a line to the `plugin.yml` so that the new option depends on the `run` property.

And while we are at it, update the version in the `README.md` file to `4.7.0` and we'll merge and make the release :)

trvrnrth:
@toote Thanks for the quick review. Updated as suggested.

toote (approved):

PR #374, opened 2023-02-20
cli-version: Fix int/string mismatch.

toothbrush:
Over in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/334, support was added for using Docker CLI v2.

However, i found it impossible to actually select!  Here's the pipeline yaml i tried:

```yaml
steps:
  - name: Build ecs-upload-task Docker
    plugins:
    - docker-compose#v4.9.0:
        cli-version: 2
        config: ecs-upload-task/docker-compose.yaml
        push: ecs-upload-task
```

The problem is though, the `plugin.yml` file specifies an "impossible schema".  Note how the field is defined as requiring a string, but the only two enum options are YAML integers!  My small change made it possible to actually use the plugin as intended and unblocked me.

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/fe3522acb645f4471014f4218fa7d884024f0b29/plugin.yml#L32-L36

Here's what i saw when i originally set `cli-version: 2` or `cli-version: "2"`.

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following pipeline will run `test.sh` inside a `app` service container using
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           run: app
 ```
 
@@ -28,7 +28,7 @@ through if you need:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           run: app
           config: docker-compose.tests.yml
           env:
@@ -41,7 +41,7 @@ or multiple config files:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           run: app
           config:
             - docker-compose.yml
@@ -56,7 +56,7 @@ env:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           run: app
 ```
 
@@ -65,7 +65,7 @@ If you want to control how your command is passed to docker-compose, you can use
 ```yml
 steps:
   - plugins:
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           run: app
           command: ["custom", "command", "values"]
 ```
@@ -79,7 +79,7 @@ steps:
   - plugins:
       - docker-login#v2.0.1:
           username: xyz
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           build: app
           image-repository: index.docker.io/myorg/myrepo
   - wait
@@ -87,7 +87,7 @@ steps:
     plugins:
       - docker-login#v2.0.1:
           username: xyz
-      - docker-compose#v4.10.0:
+      - docker-compose#v4.10.1:
           run: app
 ```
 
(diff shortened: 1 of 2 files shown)


toothbrush:
In case someone stumbles across this before the fix is merged, here was the workaround i found:

```yaml
steps:
  - name: Build ecs-upload-task Docker
    plugins:
    - docker-compose#v4.9.0:
        push:
          - ecs-upload-task
    env:
      BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLI_VERSION: "2"
```

plugin.yml
```diff
@@ -32,8 +32,8 @@ configuration:
     cli-version:
       type: string
       enum:
-        - 1
-        - 2
+        - "1"
+        - "2"
```
toote, line 36: How about allowing both versions?

```suggestion
      oneOf:
      - type: string
        enum: [ "1", "2" ]
      - type: integer
        enum: [ 1, 2 ]
```

README.md
```diff
@@ -650,9 +650,9 @@ Select when to upload container logs.
 
 The default is `on-error`.
 
-### `cli-version` (optional)
+### `cli-version` (optional, string)
 
-If set to `2`, plugin will use `docker compose` to execute commands; otherwise it will default to version `1` using `docker-compose` instead.
+If set to `"2"`, plugin will use `docker compose` to execute commands; otherwise it will default to version `"1"` using `docker-compose` instead.
```
toote, line 655: If you take my other suggestion, this change is no longer needed

README.md
```diff
@@ -650,9 +650,9 @@ Select when to upload container logs.
 
 The default is `on-error`.
 
-### `cli-version` (optional)
+### `cli-version` (optional, string)
```
toote, line 653: If you take my other suggestion this should be changed to:

```suggestion
### `cli-version` (optional, string or integer)
```

toote (commented):
This is an incredible catch! :facepalm: Just a minor suggestion to allow for both versions of the configuration. What do you think?

plugin.yml
toothbrush: Hey @toote, thanks for your feedback.  

My personal taste leans towards unambiguous and clear input requirements, so if it were me i'd pick _either_ a string or integer type for this field.  The reason i picked string (even though the version currently is a number – although in future i could imagine "2.1" being a valid version) is because of the symmetry with being able to provide arguments as environment variables (in this case `BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLI_VERSION`) – environment variables only support string types.  So to avoid a weird impedance mismatch there, i picked string.

However, i totally understand that you may have different trade-offs in mind (such as user friendliness or a different philosophy about strict input validation), and it's your project, so i'm more than happy to adapt my patch to what you decide. 🙌

toothbrush: I just went and double-checked, and at least it looks like all the values end up cast to strings anyway so at least we wouldn't have to do extra work to support your integer-or-string schema. 👍 

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/fe3522acb645f4471014f4218fa7d884024f0b29/lib/shared.bash#L37-L42

toothbrush: I've been thinking about it, and after all i think your suggestion is a good one.  I know how much of a pain YAML is with strings vs integers and so on, and it probably can't hurt to support both.  I'll make the changes :+1:

toothbrush:
> What do you think?

I've taken your suggestion – thanks @toote.

toothbrush:
image

Um, that wasn't my intention at all when clicking the re-request button!  Sorry!

toote (approved):

toothbrush:
Thanks Matías! 🙌

PR #376, opened 2023-02-27
Add labels to run container

francoiscampbell:
Adds labels that can be useful for finding the containers spawned on a host by this plugin. This can also enable automated tools to associate these docker-compose containers with the agent that started them.

Tested on our infra:
image

Question for maintainers: should this be behind a plugin config? I don't think so since it doesn't affect the behaviour of the container itself, but I'm happy to make it conditional if needed.

### DIFF
--- a/tests/push.bats
+++ b/tests/push.bats
@@ -8,10 +8,6 @@ load '../lib/shared'
 # export BUILDKITE_AGENT_STUB_DEBUG=/dev/tty
 # export BATS_MOCK_TMPDIR=$PWD
 
-setup_file() {
-  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_LABELS="false"
-}
-
 @test "Push a single service with an image in it's config" {
   export BUILDKITE_JOB_ID=1111
   export BUILDKITE_PLUGIN_DOCKER_COMPOSE_PUSH=app
--- a/tests/push.bats
+++ b/tests/push.bats
@@ -8,6 +8,10 @@ load '../lib/shared'
 # export BUILDKITE_AGENT_STUB_DEBUG=/dev/tty
 # export BATS_MOCK_TMPDIR=$PWD
 
+setup_file() {
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_LABELS="false"
+}
+
 @test "Push a single service with an image in it's config" {
   export BUILDKITE_JOB_ID=1111
   export BUILDKITE_PLUGIN_DOCKER_COMPOSE_PUSH=app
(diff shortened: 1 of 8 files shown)


toote:
This is a great addition! :clap:

I would indeed add it behind a configuration, but have it on by default... the reason is two-fold:
1- whoever doesn't want/need the behaviour will be able to turn it off
2- it will make it simpler to not have to change all 75 tests and only turn it on in the feature-specific tests :wink: :wink:

francoiscampbell:
Thanks for the feedback @toote, made it conditional (default `true`), updated the existing tests to disable the option, and added a new test.

tests/push.bats
```diff
@@ -8,6 +8,10 @@ load '../lib/shared'
 # export BUILDKITE_AGENT_STUB_DEBUG=/dev/tty
 # export BATS_MOCK_TMPDIR=$PWD
 
+setup_file() {
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_LABELS="false"
+}
+
```
toote, line 14: while appreciated, this is not actually needed as the behaviour being tested here does not change :)

tests/run.bats
```diff
@@ -1498,4 +1502,43 @@ export BUILDKITE_JOB_ID=1111
...
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN=myservice
+  export BUILDKITE_PIPELINE_SLUG=test
+  export BUILDKITE_BUILD_NUMBER=1
+  export BUILDKITE_COMMAND=pwd
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_LABELS="true"
+  export BUILDKITE_PIPELINE_NAME="label-test"
+  export BUILDKITE_LABEL="Testjob"
+  export BUILDKITE_STEP_KEY="test-job"
+  export BUILDKITE_AGENT_NAME="agent"
+  export BUILDKITE_AGENT_ID="1234"
```
toote, line 1518: If I'm being quite picky, I'd rather these variables be split in two groups: those general to the pipeline and those that are specific to the plugin. And those groups be alphabetically sorted while we are at it :P (feel free to ignore this completely cosmetic and personal opinion)

toote (approved):
:clap: @francoiscampbell 

Just a minor comment and a cosmetic suggestion.

If you feel like making the changes, I would also suggest updating the plugin's versions in the readme with the next one (`v4.11.0`) to avoid having to create another PR in the future. But there is no hurry or need and we might merge and release this plugin without that change

tests/run.bats
francoiscampbell: updated

pzeballos (approved):
Nice!

PR #384, opened 2023-04-07
Add skip-pull support to build phase

supaspoida:
Reuses the existing `skip-pull` option to conditionally remove the `--pull` flag from the generated docker compose command.

I have a use case where I am pre-building images outside the docker-compose plugin and wanted to use those images in the `FROM` tag of a `Dockerfile` to be built via the plugin. With the `--pull` flag docker compose tries to fetch from docker hub and fails. This change puts the flag behind the same `SKIP_PULL` conditional currently used by
the run phase.

I verified that this works as expected for my own pipelines but those are very simple at the moment and not exercising much of the plugin's surface area. If there are explicit reasons why this was unsupported previously I'm happy to revisit this approach or come up with another workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,7 +486,7 @@ Whether to skip the repository checkout phase. This is useful for steps that use
 
 **Important**: as the code repository will not be available in the step, you need to ensure that the docker compose file(s) are present in some way (like using artifacts)
 
-### `skip-pull` (optional, run only)
+### `skip-pull` (optional, build and run only)
 
 Completely avoid running any `pull` command. Images being used will need to be present in the machine from before or have been built in the same step. Could be useful to avoid hitting rate limits when you can be sure the operation is unnecessary. Note that it is possible other commands run in the plugin's lifecycle will trigger a pull of necessary images.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -486,7 +486,7 @@ Whether to skip the repository checkout phase. This is useful for steps that use
 
 **Important**: as the code repository will not be available in the step, you need to ensure that the docker compose file(s) are present in some way (like using artifacts)
 
-### `skip-pull` (optional, run only)
+### `skip-pull` (optional)
```
toote, line 489: It should be build and run only (as it wouldn't apply to the `push`)

toote (changes requested):
Great work!

Only one thing missing: [adding `build` to the dependency for the new option](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/51bba98ea94a130af4d84ce404c571484bc85c85/plugin.yml#L144)

README.md
supaspoida: Ah thanks, fixed in latest!

toote (approved):

PR #386, opened 2023-05-04
Remove orphaned containers in cleanup

raylu:
When I cancel a job, the main/run container continues running (but the output doesn't show up in the web interface).
See #389.

### DIFF
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -18,9 +18,9 @@ compose_cleanup() {
 
   # Stop and remove all the linked services and network
   if [[ "$(plugin_read_config LEAVE_VOLUMES 'false')" == "false" ]]; then
-    run_docker_compose down --volumes || true
+    run_docker_compose down --remove-orphans --volumes || true
   else
-    run_docker_compose down || true
+    run_docker_compose down --remove-orphans || true
   fi
 }
 
@@ -72,7 +72,7 @@ check_linked_containers_and_save_logs() {
 }
 
 # docker-compose's -v arguments don't do local path expansion like the .yml
-# versions do. So we add very simple support, for the common and basic case.
+# versions do. So we add very simple support for the common and basic case.
 #
 # "./foo:/foo" => "/buildkite/builds/.../foo:/foo"
 expand_relative_volume_path() {
--- a/tests/docker-compose-cleanup.bats
+++ b/tests/docker-compose-cleanup.bats
@@ -13,7 +13,7 @@ load '../lib/run'
   assert_success
   assert_equal "${lines[0]}" "kill"
   assert_equal "${lines[1]}" "rm --force -v"
-  assert_equal "${lines[2]}" "down --volumes"
+  assert_equal "${lines[2]}" "down --remove-orphans --volumes"
 }
 
 @test "Possible to gracefully shutdown containers in docker-compose cleanup" {
@@ -26,7 +26,7 @@ load '../lib/run'
   assert_success
   assert_equal "${lines[0]}" "stop"
   assert_equal "${lines[1]}" "rm --force -v"
-  assert_equal "${lines[2]}" "down --volumes"
+  assert_equal "${lines[2]}" "down --remove-orphans --volumes"
 }
 
 @test "Possible to skip volume destruction in docker-compose cleanup" {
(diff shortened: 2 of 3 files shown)


toote:
@raylu your situation to need this is very weird as the main container should have stopped before it gets to that point. 

First, when the job was cancelled as the agent running the job will send the whole process stack a stopping and then a termination signal. The main container should have received that and stopped itself then. If that was not enough, the code just above the lines you added should have also taken care of it by killing all containers associated to the project itself (the main container is part of the project as well). There is something definitely fishy going on that we've never seen before that I don't think should need special accommodation from the plugin, but if you create an issue in the repository we'll be more than happy to help you troubleshoot.

tests/docker-compose-cleanup.bats
```diff
@@ -4,19 +4,8 @@ load '/usr/local/lib/bats/load.bash'
...
-    echo "$@"
-  }
-  run compose_cleanup
-
-  assert_success
-  assert_equal "${lines[0]}" "kill"
-  assert_equal "${lines[1]}" "rm --force -v"
-  assert_equal "${lines[2]}" "down --volumes"
-}
-
```
toote, line 18: why did you remove this test?

lib/run.bash
toote, line 1: I had a lot of comments about how to make your code better basically on every single line you changed... until I found out that it there is a 2-line change that makes everything better: adding `--remove-orphans` to the calls to `run_docker_compose down`. Would you mind implementing and testing that instead?

toote (changes requested):
2 (big) comments. Thanks a lot for your time and patience!

lib/run.bash
raylu: if the machine happens to be running another container (from another pipeline with a different `docker-compose.yml`, for example), will `--remove-orphans` rm those too?

tests/docker-compose-cleanup.bats
raylu: maybe I misunderstood, but I felt this behavior was already covered by `Cleanup runs after a run command` in `cleanup.bats`

lib/run.bash
toote: from my tests, it would only remove the containers in the current project (I tested it on a docker host running other containers and they were not affected)

tests/docker-compose-cleanup.bats
toote: There is a lot of overlap in tests... but the one in the `cleanup.bats` file test the `pre-exit` hook script, these ones test the specific functions in the `lib/shared.bash` file instead. If the latter break, it is almost sure that the hook script that uses it will break, but the inverse is not true

raylu:
with this new code, I'm getting
```
$ docker compose -f docker-compose.buildkite.yml -p buildkite0188080e66054253ab6ce4b3e3877dec down --remove-orphans --volumes
--
  | [+] Running 0/1
  | ⠿ Container buildkite0188080e66054253ab6ce4b3e3877dec_tests_build_287311  Error while Removing 0.7s
  | Error response from daemon: removal of container 2cbf7301d12f1cac79ca2e69da1d494d7b2d1c2863222487b3b5470d1695d455 is already in progress
```
I'll report back if it actually worked in a bit. please hold off on merging

toote (approved):
👏

toote:
> I'll report back if it actually worked in a bit. please hold off on merging

I read your message after merging :( I have replied to your findings in the related Issue so that we can continue the discussion there and I'll hold-off on releasing the merged changes until we are sure it is working properly

raylu:
no worries! my fault for pushing here before testing

it appears to basically work. of our 100 test shards, 99 didn't send any reports. 1 of them did. it logged

```
Cleaning up after docker-compose | 10s
-- | --
  | $ docker compose -f docker-compose.buildkite.yml -p buildkite0188080e6624499ebe42c0980be563da kill
  | [+] Running 0/5
  | ⠴ Container buildkite0188080e6624499ebe42c0980be563da-[redactedA]-1  Killing 9.6s
  | ⠴ Container buildkite0188080e6624499ebe42c0980be563da-[redactedB]-1  Killing 9.6s
  | ⠴ Container buildkite0188080e6624499ebe42c0980be563da-[redactedC]-1  Killing 9.6s
  | ⠴ Container buildkite0188080e6624499ebe42c0980be563da-[redactedD]-1  Killing 9.6s
  | ⠴ Container buildkite0188080e6624499ebe42c0980be563da-[redactedE]-1  Killing 9.6s
```

and no further logs (so it never got to `down --remove-orphans`). this is similar to what I saw with my old version of the code using `stop $(docker ps --filter=...)`: most shards stopped the container but some leaked through. this is still a big improvement, so I'm looking forward to a release

PR #387, opened 2023-05-05
Introduce quiet-pull to avoid huge amount of junk logs get printed as CI logs  

Songyu-Wang:
I currently have a buildkite pipeline that generates 30mb to 60mb logs because this plugin prints out everything from docker-compose.

This is actually the first time I am touching this codebase. All I did was pattern matching. Obviously, I don't have a buildkite account that I can freely install plugins on to test my code, so please double-check if I made any mistakes. 

Also, this is just me trying this out and see if I can actually get the logs reduced. If this works, I may come back and implement this for all other possible commands. TBD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,7 +636,8 @@ The default is `false`.
 
 ### `quiet-pull` (optional, run only)
 
-Sets `docker-compose up` to run with `--quiet-pull`. There are other commands that can accept `--quiet-pull` or equivalents. They have not been implemented yet. 
+
+Start up dependencies with `--quiet-pull` to prevent even more logs during that portion of the execution.
 
 The default is `false`.
 
--- a/README.md
+++ b/README.md
@@ -634,6 +634,12 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
+### `quiet-pull` (optional, run only)
+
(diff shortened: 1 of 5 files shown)


Songyu-Wang:
Sorry for the ping. It seems like you are one of the admins for the repo. Just wanna bring this to your attention. Thanks!
@pzeballos

pzeballos:
No problem @Songyu-Wang! Thanks for your contributions! We'll check this out as soon as possible 🙏🏻

README.md
```diff
@@ -634,6 +634,12 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
+### `quiet-pull` (optional, run only)
+
+Sets `docker-compose up` to run with `--quiet-pull`. There are other commands that can accept `--quiet-pull` or equivalents. They have not been implemented yet. 
+
+The default is `false`.
```
toote, line 641: As the `up` command is only used when starting dependencies, I would re-word this based on that and remove the additional clarification:

```suggestion

Start up dependencies with `--quiet-pull` to prevent even more logs during that portion of the execution.

The default is `false`.
```

toote (changes requested):
This is a great idea and a very straightforward implementation 👏 

In addition to my comment about documentation, there are missing tests for the old version of docker (`tests/run.bats`)

Songyu-Wang:
@toote Thanks!

toote (approved):

PR #388, opened 2023-05-05
`env-propagation-list`: programmatically propagate envvars to the container

amartani:
Implements the same feature as https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/238, which is also useful for docker-compose use case. See the linked PR for rationale / use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ If you set this to `VALUE`, and `VALUE` is an environment variable containing a 
 
 Whether or not to automatically propagate all pipeline environment variables into the run container. Avoiding the need to be specified with environment.
 
-\* Caveat: only environment variables listed in $BUILDKITE_ENV_FILE will be propagated. This does not include e.g. variables that you exported in an `environment` hook. If you wish for those to be propagated, try `env-propagation-list`.
+**Important**: only pipeline environment variables will be propagated (what you see in the BuildKite UI, those listed in `$BUILDKITE_ENV_FILE`). This does not include variables exported in preceeding `environment` hooks. If you wish for those to be propagated you will need to list them specifically or use `env-propagation-list`.
 
 ### `command` (optional, run only, array)
 
--- a/README.md
+++ b/README.md
@@ -460,11 +460,15 @@ A list of KEY=VALUE that are passed through as build arguments when image is bei
 
 A list of either KEY or KEY=VALUE that are passed through as environment variables to the container.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -460,11 +460,15 @@ A list of KEY=VALUE that are passed through as build arguments when image is bei
...
+### `env-propagation-list` (optional, string)
+
+If you set this to `VALUE`, and `VALUE` is an environment variable containing a space-separated list of environment variables such as `A B C D`, then A, B, C, and D will all be propagated to the container. This is helpful when you've set up an `environment` hook to export secrets as environment variables, and you'd also like to programmatically ensure that secrets get propagated to containers, instead of listing them all out.
+
 ### `propagate-environment` (optional, boolean)
 
 Whether or not to automatically propagate all pipeline environment variables into the run container. Avoiding the need to be specified with environment.
 
-**Important**: only pipeline variables will automatically be propagated (what you see in the Buildkite UI). Variables set in proceeding hook scripts will not be propagated to the container.
+\* Caveat: only environment variables listed in $BUILDKITE_ENV_FILE will be propagated. This does not include e.g. variables that you exported in an `environment` hook. If you wish for those to be propagated, try `env-propagation-list`.
```
toote, line 471: I'm not sure that everyone will understsand the file reference compare to what it was before, but I do like the rest of the rewording. How about something like:

```suggestion
**Important**: only pipeline environment variables will be propagated (what you see in the BuildKite UI, those listed in `$BUILDKITE_ENV_FILE`). This does not include variables exported in preceeding `environment` hooks. If you wish for those to be propagated you will need to list them specifically or use `env-propagation-list`.
```

toote (changes requested):
Just a minor comment on the documentation, but also the tests should be added when compose is run as a docker plugin instead of a standalone executable (in the file `tests/v2/run.bats`)

README.md
amartani: I had copied the wording of the original PR (https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/238) since the effect of the variable is the same, but I agree that the extra explanation seems better. Thanks!

amartani:
> Just a minor comment on the documentation, but also the tests should be added when compose is run as a docker plugin instead of a standalone executable (in the file `tests/v2/run.bats`)

Added tests for tests/v2/run.bats; PTAL.

toote (approved):

PR #401, opened 2023-07-04
Default tty to false to minimise build output

chendo:
This PR changes the default of `tty` to `false`.

Unless there's a scenario I'm not considering, it doesn't seem to make sense to have this on by default, as CI is traditionally a non-interactive environment anyway. Most command line tools change their output behaviour to be more "interactive" when a TTY is detected, but this drastically increases build output, causing build log truncation warnings.

For a single step in our pipeline, the build output size difference is staggering!

* `tty: true` - 2.1MB
* `tty: false` - 752KB (35% of above output, ~65% reduction)

### DIFF
--- a/README.md
+++ b/README.md
@@ -600,9 +600,9 @@ The default is `false`.
 
 ### `tty` (optional, run only)
 
-If set to false, doesn't allocate a TTY. This is useful in some situations where TTY's aren't supported, for instance windows.
+If set to true, allocates a TTY. This is useful in some situations TTYs are required.
 
-The default is `true` on unix, `false` on windows
+The default is `false`.
 
 ### `dependencies` (optional, run only)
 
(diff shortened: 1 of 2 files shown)


pzeballos:
Hey @chendo! thanks! I believe it's a good idea, but it would require a major release which we are planning to do in the following months to make docker compose v2 the default one.
This PR will also need to update the tests 🙃 (they are currently failing)

chendo:
I had a quick look at the tests but it's unclear how it works and unfortunately I won't have time to address those

pzeballos:
No problem! We can work on a new PR with all the tests once we are ready for the v5 of the plugin (which should be during Q4)
I'll close this one, but we'll mention you once it's ready.
Thanks again! 🙏🏻

PR #407, opened 2023-10-12
Collapse logging of run output by default

francoiscampbell:
By default this plugin collects the run output in a `+++` group, while the `docker` plugin collects it in a `---` group (https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/36b85cc317745f868f2f6918a02f9cfb6a0b3069/commands/run.sh#L503). This PR unifies the two plugins by switching to using `---` by default.

The rationale for using `---` by default is that when we're passing lots of environment variables, volumes, etc, the `docker-compose run ...` command gets pretty long and having the log group expanded by default is noisy. Additionally, it's quite likely that the command being run will create a `+++` group of its own, so it's not necessary for the plugin to provide one by default.

For backwards compatibility, the `collapse-run-log-group` option has been added, which will restore the old behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following pipeline will run `test.sh` inside a `app` service container using
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.14.0:
+      - docker-compose#v4.15.0:
           run: app
 ```
 
@@ -28,7 +28,7 @@ through if you need:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.14.0:
+      - docker-compose#v4.15.0:
           run: app
           config: docker-compose.tests.yml
           env:
@@ -41,7 +41,7 @@ or multiple config files:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.14.0:
+      - docker-compose#v4.15.0:
           run: app
           config:
             - docker-compose.yml
@@ -56,7 +56,7 @@ env:
 steps:
   - command: test.sh
     plugins:
-      - docker-compose#v4.14.0:
+      - docker-compose#v4.15.0:
           run: app
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -430,6 +430,14 @@ Will cause the image to be pushed twice (once by the build step and another by t
 
 Pull down multiple pre-built images. By default only the service that is being run will be pulled down, but this allows multiple images to be specified to handle prebuilt dependent images. Note that pulling will be skipped if the `skip-pull` option is activated.
 
+### `collapse-run-log-group` (optional, boolean, run only)
+
+Whether to collapse or expand the log group that is created for the output of `docker-compose run`. When this setting is `true`, the output is collected into a `---` group, when `false` the output is collected into a `+++` group. Setting this to `false` can be useful to highlight your command's output if it does not create its own `+++` group.
+
+For more information see [Managing log output](https://buildkite.com/docs/pipelines/managing-log-output).
+
+Default `true`
```
pzeballos, line 439: For backwards compatibility this should be `false` by default

pzeballos (changes requested):
Thanks @francoiscampbell! I think this is reasonable, but you are changing the default behavior, if the default kept the current functionality will be ok to merge (and update the test 🙂 )

francoiscampbell:
Ok @pzeballos, I've flipped the default. Are we OK with the default being inconsistent between this plugin and the Docker one?

pzeballos:
> Ok @pzeballos, I've flipped the default. Are we OK with the default being inconsistent between this plugin and the Docker one?

yep, all good, they are different plugins, and we prefer not to do a backwards compatibility change at the moment :) 

The build is failing in the linter step @francoiscampbell , can you update the Readme to use v4.15.0? Thanks!

francoiscampbell:
Updated

pzeballos (approved):
Thanks!

PR #408, opened 2023-10-12
Add --skip-pull support to v2 run command

eugeneotto:
The `docker-compose` plugin currently seems to ignore the `skip-pull` option. For example, this configuration:

```
steps:
  - command: bundle exec rubocop
    plugins:
      - docker-compose#v4.14.0:
          run: ruby
          skip-pull: true
          config: docker-compose.yml
          cli-version: 2
```

Results in a command that includes the `--pull` flag, something like this:

```
docker compose -f docker-compose.yml -p buildkite0123456789 build --pull ruby
```

This PR updates run.sh to fully support the `skip-pull` option. The same configuration as above now results in the `--pull` flag being omitted from the generated command:

```
docker compose -f docker-compose.yml -p buildkite0123456789 build ruby
```

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -271,7 +271,12 @@ fi
 
 run_params+=("$run_service")
 
-build_params=(--pull)
+build_params=()
+
+# Only pull if SKIP_PULL is not true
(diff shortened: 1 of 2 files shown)


pzeballos (approved):
looks good! thanks!

Issue #391, opened 2023-05-18
The `--target` option doesn't seem to exist


viraptor:
I think there was a mistake where the `target` option was added to the plugin. Specifically in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/231 the referenced documentation is https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage

That option is specific to `docker build` though, not `docker-compose build`. If I understand the original request correctly, the way to do this in the plugin would be to add the option to the override yml instead.

Currently the plugin tries to add `--target ...` to the `docker-compose build` command and fails because it's not implemented https://docs.docker.com/engine/reference/commandline/compose_build/

(Unless this is some version specific behaviour that was added/removed in the meantime)

viraptor:
> I am really interested in your expectations on how it should work and what kind of configurations you want off it.

My original use case was to have a service `app` pointing at a Dockerfile with multiple stages. Then the test steps would use an override to use `target: ci` which includes the test tools, while other steps would use the default which builds just the deployable image.

Turns out I had more differences to implement, so ended up just duplicating the service as `...-prod` and `...-test` in the docker-compose itself and assigning different targets there.

I think the option could still be useful in a simpler setup.

> Also, I am not 100% sure how specifying a target would work when you are building multiple services. Do you have any thoughts on that?

Nope, I suspect it would have to be a hack for a single service only.

camallen:
I just ran into this issue and assumed that specifying the `target` option would override the compost file to add a target directive to my service, thus allowing me to target a stage in a multi stage build. Sadly this just slaps an incorrect `--target` onto the docker compose cmd which results in 
```
$ docker-compose -f docker-compose.yml -p buildkite018c900423954192bc6a0b21321ae8ae -f docker-compose.buildkite-10-override.yml build --pull --target base app
unknown flag: --target
🚨 Error: The command exited with status 16
```

FWIW - My use case is similar to @viraptor in that i want to target a multi stage build. 

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/231#issuecomment-550067960 provides a similar use case I have where i'd like to prebuild multi stage cache images to avoid extra work in subsequent builds.

PR #418, opened 2024-01-10
Build simplification

toote:
Simplified configuration removing `image-repository` and `image-name` configurations. Changed the way build and push interact. Now builds do not push, so metadata to signal pre-built images to further steps in the pipeline is moved to pushes as well.

Also refactored the logic to create override files only when necessary and with other information (like `target` values), thus correcting #391

### DIFF
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -12,6 +12,11 @@ setup_file() {
   export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_LABELS="false"
 }
 
+teardown() {
+  # some test failures may leave this file around
+  rm -f docker-compose.buildkite-1-override.yml
+}
+
 @test "Run without a prebuilt image" {
   export BUILDKITE_JOB_ID=1111
   export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN=myservice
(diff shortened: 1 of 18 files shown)


tests/run.bats
```diff
@@ -12,6 +12,11 @@ setup_file() {
   export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_LABELS="false"
 }
 
+teardown() {
+  # some test failures may leave this file around
+  rm -f docker-compose.buildkite-1-override.yml
```
tomowatt, line 17: Don't believe it will make much of a difference other than for consistency but noticed in `tests/v2/run.bats` is using a glob
```
rm -f docker-compose.buildkite*-override.yml
```

tomowatt (approved):
All looks good, re-ran the tests and getting expected results

zzak:
Just FYI, I got tripped up on this, removing `image-name` as a way to specify a tag to push being abrubtly removed was startling and the upgrade docs wasn't very clear about why.

Maybe I misread "no longer necessary" as "redundant" and not "needs replacement".

Not sure there is anything actionable but wanted to pass along the feedback. :bow:

PR #420, opened 2024-01-10
CLI v2 as default

### LINKED ISSUE #390: `docker-compose` reaching end of support, migrate to `docker compose`
According to this https://docs.docker.com/compose/migrate/ , Compose V1 is reaching end of support. In fact, a warning at the bottom of this page indicates they had even stopped security updates back in 2021. The recommendation is to migrate to using the built in docker version of compose (aka "v2"). Generally the change is to just go from `docker-compose` based commands to `docker compose` (no hyphen) based commands.

toote:
:scream:

Updates the whole plugin to use CLI v2 as the default instead of v1 that has been deprecated.

Closes #390

### DIFF
--- a/tests/v1/build.bats
+++ b/tests/v1/build.bats
@@ -3,8 +3,12 @@
 load "${BATS_PLUGIN_PATH}/load.bash"
 load '../../lib/shared.bash'
 
- export DOCKER_COMPOSE_STUB_DEBUG=/dev/stdout
+# export DOCKER_COMPOSE_STUB_DEBUG=/dev/stdout
 # export BATS_MOCK_TMPDIR=$PWD
+
+setup_file() {
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLI_VERSION=1
+}
 
 teardown() {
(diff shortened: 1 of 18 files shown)


tests/v1/build.bats
```diff
@@ -1,19 +1,22 @@
...
 setup_file() {
-  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLI_VERSION=2
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLI_VERSION=1
 }
 
 teardown() {
-  # plugin leaves override files around
-  rm -f docker-compose.buildkite*-override.yml
+  # some test failures may leave this file around
+  rm -f docker-compose.buildkite-1-override.yml
```
tomowatt, line 15: I'm sure this is change is reverted and corrected in another PR, but highlighting it

toote: Good catch... applied it again because I had to resolve a few conflicts anyways

tomowatt (approved):
All looks good to me and happy with the testing and backwards compatibility

PR #421, opened 2024-01-11
Documentation updates

toote:
* Splits Readme into 2 files: one with options, another with examples
* Rewording of all options
* review of plugin schema to ensure all options are properly validated
* update of the pipeline to validate examples in the new file
* bump version to 5.0.0 in preparation of upcoming major release
* adds a new document with what I could think may help users migrate to this new major version


plugin.yml
```diff
@@ -14,6 +14,7 @@ configuration:
     push:
       type: [ string, array ]
       minimum: 1
+
```
tomowatt, line 17: Checking this additional line is expected to break up the plugin "commands" from the additional arguements?

tomowatt (approved):
Looks good to me, only a comment but not a blocker

plugin.yml
toote: I added it on purpose to separate the _"main commands"_ from the rest of the options that are sorted alphabetically

README.md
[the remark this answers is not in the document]
toote: good catch, I went through all options and made sure that:
* all options are documented
* all documented options are in the plugin's schema
* all options that are not just a string has valid types in the documentation and schema

PR #426, opened 2024-02-21
feat: Add option to automatically mount AWS auth tokens when using docker compose run command

christophe-scalepad:
Like the docker plugin, it can be useful to automatically mount AWS auth token from the host to the container when using the docker compose plugin


pzeballos (approved):
Thanks @christophe-scalepad!

PR #428, opened 2024-02-23
Add buildkite inline cache env

jquick:
It's possible to set this via the args config property. However, when trying to define it as global default with
```
BUILDKITE_PLUGIN_DOCKER_COMPOSE_ARGS_0="BUILDKIT_INLINE_CACHE=1"
```
you can end up accidentally overwriting args provided in a pipeline.yml.

Providing a separate config property allows you to set it via an env var
while avoiding overwriting any provided args.

### Use Case
We build a lot of images and usually have our users setup their own pipelines. As owners we want to ensure that every image is built with inline cache.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,12 @@ If set to `true` it will mount onto `/workdir`, unless `workdir` is set, in whic
 
 Default: `false`
 
+### `buildkit-inline-cache` (optional, build-only, boolean)
+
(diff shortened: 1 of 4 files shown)


toote (approved):
:clap: how did we not think of this? :clap:

pzeballos (approved):
Thanks @jquick! 🙌🏻 this looks very good! and thanks for fixing the tests :P

PR #429, opened 2024-02-23
Add service label build options

jquick:
It would be really handy to be able to add service labels in the pipeline. This change allows labels to be defined for build or run.

This will add any defined labels to the compose override and those will be merged accordingly. Small example:

docker-compose.yml
```
services:
  app:
    build:
      context: .
      labels:
        - dev.pardot.docker.image-metadata=test-metadata
```

pipeline.yml
```
      - ${DOCKER_COMPOSE_PLUGIN}:
          build: app
          labels: 
            - dev.pardot.docker.test=test
            - dev.pardot.docker.test2=test2
```

image labels after build:
```
      "Labels": {
        "com.docker.compose.project": "buildkite018dd7eac0554b8eb556b2b206b874d0",
        "com.docker.compose.service": "app",
        "com.docker.compose.version": "2.12.2",
        "dev.pardot.docker.image-metadata": "test-metadata",
        "dev.pardot.docker.test": "test",
        "dev.pardot.docker.test2": "test2"
      }
```

### Testing

I've done some basic Buldkite builds with different combinations of `labels` and `cache_from`.

### Use Case
Our users setup their own compose files. As owners we want to ensure that every image is built with specific labels.

### DIFF
--- a/lib/shared.bash
+++ b/lib/shared.bash
@@ -170,6 +170,7 @@ function build_image_override_file_with_version() {
     cache_from_amt="${1:-0}"
     [[ -n "${1:-}" ]] && shift; # remove the value if not empty
     if [[ "${cache_from_amt}" -gt 0 ]]; then
+      cache_from=()
       for _ in $(seq 1 "$cache_from_amt"); do
         cache_from+=( "$1" ); shift
       done
@@ -179,6 +180,7 @@ function build_image_override_file_with_version() {
     labels_amt="${1:-0}"
     [[ -n "${1:-}" ]] && shift; # remove the value if not empty
     if [[ "${labels_amt}" -gt 0 ]]; then
+      labels=()
       for _ in $(seq 1 "$labels_amt"); do
         labels+=( "$1" ); shift
       done
--- a/lib/shared.bash
+++ b/lib/shared.bash
@@ -170,7 +170,7 @@ function build_image_override_file_with_version() {
     cache_from_amt="${1:-0}"
     [[ -n "${1:-}" ]] && shift; # remove the value if not empty
     if [[ "${cache_from_amt}" -gt 0 ]]; then
-      for amt in $(seq 1 "$cache_from_amt"); do
+      for _ in $(seq 1 "$cache_from_amt"); do
         cache_from+=( "$1" ); shift
       done
     fi
@@ -179,7 +179,7 @@ function build_image_override_file_with_version() {
     labels_amt="${1:-0}"
     [[ -n "${1:-}" ]] && shift; # remove the value if not empty
     if [[ "${labels_amt}" -gt 0 ]]; then
-      for amt in $(seq 1 "$labels_amt"); do
+      for _ in $(seq 1 "$labels_amt"); do
         labels+=( "$1" ); shift
       done
     fi
--- a/lib/shared.bash
+++ b/lib/shared.bash
@@ -164,10 +164,29 @@ function build_image_override_file_with_version() {
     service_name=$1
     image_name=$2
     target=$3
-    cache_from_amt=$4
-    shift 4
+    shift 3
 
-    if [[ -z "$image_name" ]] && [[ -z "$target" ]] && [[ "$cache_from_amt" -eq 0 ]]; then
+    # load cache_from array
+    cache_from=()
+    if (( $# > 0 )); then
+      cache_from_amt=$1
+      shift
+      while (( cache_from_amt-- > 0 )) ; do
+        cache_from+=( "$1" ); shift
+      done
+    fi
+
+    # load labels array
+    labels=()
+    if (( $# > 0 )); then
+      labels_amt=$1
+      shift
+      while (( labels_amt-- > 0 )) ; do
+        labels+=( "$1" ); shift
+      done
+    fi
+
+    if [[ -z "$image_name" ]] && [[ -z "$target" ]] && [[ "${#cache_from[@]}" -eq 0 ]] && [[ "${#labels[@]}" -eq 0 ]]; then
       # should not print out an empty service
       continue
     fi
@@ -178,7 +197,7 @@ function build_image_override_file_with_version() {
       printf "    image: %s\\n" "$image_name"
     fi
 
-    if [[ "$cache_from_amt" -gt 0 ]] || [[ -n "$target" ]]; then
+    if [[ "${#cache_from[@]}" -gt 0 ]] || [[ -n "$target" ]] || [[ "${#labels[@]}" -gt 0 ]]; then
       printf "    build:\\n"
     fi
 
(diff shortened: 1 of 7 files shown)


lib/shared.bash
```diff
@@ -164,10 +164,29 @@ function build_image_override_file_with_version() {
...
+
+    # load cache_from array
+    cache_from=()
+    if (( $# > 0 )); then
+      cache_from_amt=$1
+      shift
+      while (( cache_from_amt-- > 0 )); do
+        cache_from+=( "$1" ); shift
+      done
+    fi
```
toote, line 177: for style consistency with the rest of the code I would re-write this to the equivalent (but untested):


```suggestion
    cache_from_amt="${1:-0}"

    if [[ "${cache_from_amt}" -gt 0 ]] ; then
      shift; # remove the value
      for amt in $(seq 1 "$cache_from_amt"); do
        cache_from+=( "$1" ); shift
      done
    fi
```

lib/shared.bash
```diff
@@ -164,10 +164,29 @@ function build_image_override_file_with_version() {
...
+
+    # load labels array
+    labels=()
+    if (( $# > 0 )); then
+      labels_amt=$1
+      shift
+      while (( labels_amt-- > 0 )); do
+        labels+=( "$1" ); shift
+      done
```
toote, line 186: same here:
for style consistency with the rest of the code I would re-write this to the equivalent (but untested):


```suggestion
    labels_amt="${1:-0}"

    if [[ "${labels_amt}" -gt 0 ]] ; then
      shift; # remove the value
      for amt in $(seq 1 "$labels_amt"); do
        labels+=( "$1" ); shift
      done
    fi
```

lib/shared.bash
```diff
@@ -164,10 +164,29 @@ function build_image_override_file_with_version() {
...
+    fi
 
-    if [[ -z "$image_name" ]] && [[ -z "$target" ]] && [[ "$cache_from_amt" -eq 0 ]]; then
+    if [[ -z "$image_name" ]] && [[ -z "$target" ]] && [[ "${#cache_from[@]}" -eq 0 ]] && [[ "${#labels[@]}" -eq 0 ]]; then
```
toote, line 189: With the previous changes I suggested, the `_amt` variables would always exist so this could be:

```suggestion
    if [[ -z "$image_name" ]] && [[ -z "$target" ]] && [[ "$cache_from_amt" -eq 0 ]] && [[ "$labels_amt" -eq 0 ]]; then
```

lib/shared.bash
```diff
@@ -178,15 +197,15 @@ function build_image_override_file_with_version() {
       printf "    image: %s\\n" "$image_name"
     fi
 
-    if [[ "$cache_from_amt" -gt 0 ]] || [[ -n "$target" ]]; then
+    if [[ "${#cache_from[@]}" -gt 0 ]] || [[ -n "$target" ]] || [[ "${#labels[@]}" -gt 0 ]]; then
```
toote, line 200: same here:
```suggestion
    if [[ "$cache_from_amt" -gt 0 ]] || [[ -n "$target" ]] || [[ "$labels_amt" -gt 0 ]]; then
```

lib/shared.bash
```diff
@@ -178,15 +197,15 @@ function build_image_override_file_with_version() {
...
       printf "    build:\\n"
     fi
 
     if [[ -n "$target" ]]; then
       printf "      target: %s\\n" "$target"
     fi
 
-    if [[ "$cache_from_amt" -gt 0 ]] ; then
+    if [[ "${#cache_from[@]}" -gt 0 ]] ; then
```
toote, line 208: and here:
```suggestion
    if [[ "$cache_from_amt" -gt 0 ]] ; then
```

lib/shared.bash
```diff
@@ -196,10 +215,16 @@ function build_image_override_file_with_version() {
...
 
       printf "      cache_from:\\n"
-      for cache_from_i in $(seq 1 "$cache_from_amt"); do
-        printf "        - %s\\n" "${!cache_from_i}"
+      for cache_from_i in "${cache_from[@]}"; do
+        printf "        - %s\\n" "${cache_from_i}"
+      done
+    fi
+
+    if [[ "${#labels[@]}" -gt 0 ]] ; then
```
toote, line 223: and here:
```suggestion
    if [[ "$labels_amt" -gt 0 ]] ; then
```

plugin.yml
```diff
@@ -150,6 +153,7 @@ configuration:
     environment: [ run ]
     expand-volume-vars: [ volumes ]
     graceful-shutdown: [ run ]
+    labels: [ build, run ]
```
toote, line 156: as far as I understand, this means that the schema will consider the usage of `labels` invalid unless you specify both `build` and `run`, that I don't think it is what you want here

tests/image-override-file.bats
```diff
@@ -29,8 +29,8 @@ EOF
   )
 
   run build_image_override_file_with_version "2.1" \
-    "myservice1" "newimage1:1.0.0" "" 0 \
-    "myservice2" "newimage2:1.0.0" "" 0
+    "myservice1" "newimage1:1.0.0" "" 0 0 \
+    "myservice2" "newimage2:1.0.0" "" 0 0
```
toote, line 33: if you implement my changes, these changes are no longer necessary

toote (changes requested):
The recommendations on `shared.bash` are mostly style and not functionality, but the schema one does pose a question on the implementation: if both `build` and `run` are used in a single step the labels would be the same and I don't think that is something useful. Thus, I would suggest splitting the option into two: `build-labels` (for `build`) and `labels` (for `run`).

While we are at it I would also suggest adding a test for the override function with labels, target and cache-from for completeness' sake (and I just noticed there is no test with just a target :fearful:)

tests/image-override-file.bats
jquick: I tried to get rid of this, but it was a little hard as we still have to shift if its "0". I think we would have to check if it exist and also if its numeric. The argument position should be static in the code so I _think_ it should be safe to only test this flow when using multiple services.

jquick:
Thanks @toote! I really appreciate your feedback and help. I think I have addressed the style issues and ran a few builds on the new logic and it looks to work as expected. Since we are splitting out run/build I think having `build-labels` and the current `run-labels` should be sufficient for most cases.

tests/image-override-file.bats
toote: in my suggestions I specifically used `${1:-0}` for the case when it was not set but I'm probably missing some corner case or logic... doesn't matter much :)

toote (approved):
:clap: as long as the pipeline passes, I believe this can be merged. Thanks a lot for the work and improvement!

pzeballos (approved):
Thanks @jquick! We'll probably do the release at the end of the week

PR #430, opened 2024-02-29
Add run-image option

jquick:
One final addition!

We have ran into this case a few times where one of the following happens:
* A image was created outside of the docker-compose-plugin but we still want to execute run commands on it
  * This is usually from a manual docker build
* Multiple images were created (with different names) for a single service. In this case the plugin will always pull the latest one created
  * This is usually from separate architecture builds which we combine later but want to test individually

For these cases we can inject an:
```
buildkite-agent meta-data set docker-compose-plugin-built-image-tag*
```
before running but I think an official option would be helpful.

This PR adds a new `run-image` option where you can pass an image and it will try to run with it.

An example:
```
  - label: ":docker: Build (multiplatform)"
    command: >-
      docker buildx build
      --target build
      --platform "linux/amd64,linux/arm64"
      --build-arg BUILDKIT_INLINE_CACHE=1
      --cache-from "${DOCKER_BUILD_BASE}/example/build:unstable-multi"
      -t "${DOCKER_BUILD_BASE}/example/build:${CI_BUILD_ID}"
      example/ 

  - label: ":golang: Test"
    command: go test -v ./...
    plugins:
      - ${DOCKER_COMPOSE_PLUGIN}:
          run: build
          run-image: ${DOCKER_BUILD_BASE}/example/build:${CI_BUILD_ID}
```

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -40,8 +40,16 @@ prebuilt_service_overrides=()
 prebuilt_services=()
 
 # We look for a prebuilt images for all the pull services and the run_service.
+prebuilt_image_override="$(plugin_read_config RUN_IMAGE)"
 for service_name in "${prebuilt_candidates[@]}" ; do
-  if prebuilt_image=$(get_prebuilt_image "$service_name") ; then
+  prebuilt_image=$(get_prebuilt_image "$service_name") || :
+
+  # override run_service prebuilt_image if set
+  if [[ -n "$prebuilt_image_override" ]] && [[ "$service_name" == "$run_service" ]] ; then
+    prebuilt_image="$prebuilt_image_override"
+  fi
+
+  if [[ -n "$prebuilt_image" ]] ; then
     echo "~~~ :docker: Found a pre-built image for $service_name"
     prebuilt_service_overrides+=("$service_name" "$prebuilt_image" "" 0 0)
     prebuilt_services+=("$service_name")
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -1280,3 +1280,28 @@ cmd3"
   assert_output --partial "env-propagation-list desired, but LIST_OF_VARS is not defined!"
   unstub buildkite-agent
 }
+
+@test "Run wit a run image override" {
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN=myservice
+  export BUILDKITE_COMMAND="echo hello world"
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_IMAGE=docker.buildkite.com/myservice:123
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CHECK_LINKED_CONTAINERS=false
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLEANUP=false
+
+  stub docker \
+    "compose -f docker-compose.yml -p buildkite1111 -f docker-compose.buildkite-1-override.yml pull myservice : echo pulled myservice" \
+    "compose -f docker-compose.yml -p buildkite1111 -f docker-compose.buildkite-1-override.yml up -d --scale myservice=0 myservice : echo ran myservice dependencies" \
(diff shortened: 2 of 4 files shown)


commands/run.sh
```diff
@@ -40,8 +40,16 @@ prebuilt_service_overrides=()
 prebuilt_services=()
 
 # We look for a prebuilt images for all the pull services and the run_service.
+prebuilt_image_override="$(plugin_read_config RUN_IMAGE)"
 for service_name in "${prebuilt_candidates[@]}" ; do
-  if prebuilt_image=$(get_prebuilt_image "$service_name") ; then
+  prebuilt_image=$(get_prebuilt_image "$service_name") || :
```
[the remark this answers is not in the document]
toote: indeed, though I prefer the usage of `true` instead of the more obscure `:`  to be more explicit:

```suggestion
  prebuilt_image=$(get_prebuilt_image "$service_name") || true
```

commands/run.sh
```diff
@@ -40,8 +40,16 @@ prebuilt_service_overrides=()
...
+
+  # override run_service prebuilt_image if set
+  if [[ -n "$prebuilt_image_override" ]] && [[ "$service_name" == "$run_service" ]] ; then
+    prebuilt_image="$prebuilt_image_override"
+  fi
+
+  if [[ -n "$prebuilt_image" ]] ; then
     echo "~~~ :docker: Found a pre-built image for $service_name"
```
toote, line 53: this message should be updated because the override image code would make it look in logs as if a pre-built image was found

toote, line 53: An alternative to this would be:

```suggestion
prebuilt_image_override="$(plugin_read_config RUN_IMAGE)"
for service_name in "${prebuilt_candidates[@]}" ; do
  if [[ "$service_name" == "$run_service" ]] && [[ -n "$prebuild_image_override" ]]; then
    echo "~~~ :docker: Overriding run image for $service_name"
    prebuilt_image="$prebuild_image_override"
  elif prebuilt_image=$(get_prebuilt_image "$service_name") ; then
     echo "~~~ :docker: Found a pre-built image for $service_name"
  fi

  if [[ -n "$prebuilt_image" ]] ; then
```

tests/run.bats
```diff
@@ -1280,3 +1280,28 @@ cmd3"
   assert_output --partial "env-propagation-list desired, but LIST_OF_VARS is not defined!"
   unstub buildkite-agent
 }
+
+@test "Run wit a run image override" {
```
toote, line 1284: just nit-picking here:
```suggestion
@test "Run with a run image override" {
```

tests/run.bats
```diff
@@ -1280,3 +1280,28 @@ cmd3"
...
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN=myservice
+  export BUILDKITE_COMMAND="echo hello world"
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_RUN_IMAGE=docker.buildkite.com/myservice:123
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CHECK_LINKED_CONTAINERS=false
+  export BUILDKITE_PLUGIN_DOCKER_COMPOSE_CLEANUP=false
```
toote, line 1289: these two variables should not be necessary

toote (changes requested):
Interesting addition!

There are a few minor comments on the code that I would have approved on as-is, but as the log message would need changing anyways I also proposed an alternative version of the code (that is unchecked and untested as I wrote it on-the fly) that I believe would reduce the diff and make the `||` unnecessary

commands/run.sh
jquick: Ah! good call

jquick:
Thanks @toote! I updated with your suggestions and it looks good!
Screenshot 2024-02-29 at 7 03 13 PM

Did some testing with and without and I am happy with it. Appreciate the feedback and guidance!

toote (approved):

PR #432, opened 2024-03-27
Add metadata key prefix to support builds with independent clusters/registries

donbobka:
In scenarios where agent clusters are isolated from each other, with each cluster responsible for building its own image and pushing it to its respective image registry, there is a potential race condition that can lead to errors during separate `run`/`push` steps. Specifically, one of the steps may attempt to pull an image from the wrong image registry due to shared metadata across all clusters.

To mitigate it and ensure that each cluster is building and running images from the correct image registry, I propose introducing a new `metadata_key_prefix` parameter. This parameter would allow users to specify a unique prefix for the metadata keys used by the plugin, effectively isolating the metadata for each agent cluster.

## example

docker-compose.yml
```
version: '3.10'
service:
  app:
    build: .
```

pipeline.yml
```
steps:
  - name: ":docker: Build staging image"
    key: "build_staging"
    plugins:
      ... docker login plugin into registry ${IMAGE_REPO_STAGING} ...
      - docker-compose#v5.2.0:
          build: app
          push: app:${IMAGE_REPO_STAGING}:${BUILD_TAG}
          metadata_key_prefix: "staging"
    agents:
      queue: staging-agents

  - name: ":docker: Build production image"
    key: "build_production"
    plugins:
      ... docker login plugin into registry ${IMAGE_REPO_PRODUCTION} ...
      - docker-compose#v5.2.0:
          build: app
          push: app:${IMAGE_REPO_PRODUCTION}:${BUILD_TAG}
          metadata_key_prefix: "production"
    agents:
      queue: staging-production

  - name: ":docker: run staging image"
    key: "run_staging"
    depends_on:
      - build_staging
    plugins:
      ... docker login plugin into registry ${IMAGE_REPO_STAGING} ...
      - docker-compose#v5.2.0:
          run: app
          metadata_key_prefix: "staging"
    agents:
      queue: staging-agents

  - name: ":docker: run production image"
    key: "run_production"
    depends_on:
      - build_production
    plugins:
      ... docker login plugin into registry ${IMAGE_REPO_PRODUCTION} ...
      - docker-compose#v5.2.0:
          run: app
          metadata_key_prefix: "production"
    agents:
      queue: staging-production
```

### DIFF
--- a/commands/push.sh
+++ b/commands/push.sh
@@ -26,6 +26,8 @@ if plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_COMPOSE_BUILD; then
   build_services=("${result[@]}")
 fi
 
+metadata_key_prefix="$(plugin_read_config METADATA_KEY_PREFIX '')"
+
 # Then we figure out what to push, and where
 for line in $(plugin_read_list PUSH) ; do
   IFS=':' read -r -a tokens <<< "$line"
@@ -37,7 +39,7 @@ for line in $(plugin_read_list PUSH) ; do
   elif in_array "${service_name}" "${build_services[@]}"; then
     echo "~~~ :docker: Service was built in this step, using that image"
     service_image="$(default_compose_image_for_service "${service_name}")"
-  elif prebuilt_image="$(get_prebuilt_image "$service_name")"; then
+  elif prebuilt_image="$(get_prebuilt_image "$metadata_key_prefix" "$service_name")"; then
     echo "~~~ :docker: Using pre-built image ${prebuilt_image}"
 
     # Only pull it down once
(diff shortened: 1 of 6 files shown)


toote (changes requested):
@donbobka thanks a lot for the contribution! This is an interesting approach to a problem I have seldom seen before. I find it interesting that it has not come up in the past but I do see some kind of need for it.

With that said, I am not too fond of the name of the option nor its implementation. What do you thing about renaming the option to `prebuilt-image-namespace` with a default value of `docker-compose-plugin`? That way the code would still behave the exact same way as it is right now if the option is not specified and the option name has no information about its actual implementation details (meaning that it can be freely changed in the future).

toote (approved):
👏

PR #438, opened 2024-05-07
Fix inconsistent heading in README.md

Pawka:
Fixing inconsistent heading of a single item in the `README.md`.


pzeballos (approved):
good catch!

PR #442, opened 2024-06-06
Add caveat for login requirement

mcncl:
There is an undocumented requirement that a `login` to a registry is performed if the user wishes to use the `push` command, else it will be skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ The name of the service the command should be run within. If the docker-compose 
 
 A list of services to push.  You can specify just the service name to push or the format `service:registry:tag` to override where the service's image is pushed to. Needless to say, the image for the service must have been built in the very same step or built and pushed previously to ensure it is available for pushing.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,8 @@ The name of the service the command should be run within. If the docker-compose
 
 A list of services to push.  You can specify just the service name to push or the format `service:registry:tag` to override where the service's image is pushed to. Needless to say, the image for the service must have been built in the very same step or built and pushed previously to ensure it is available for pushing.
 
+:warning: The `push` command will be skipped if a `login` command has not previously been used, or the agent has not been authenticated for a image registry.
```
pzeballos, line 35:
```suggestion
:warning: The `push` command will fail if a `login` command has not previously been used or the agent has not been authenticated for an image registry.
```

pzeballos (approved):
Nice clarification 😎. AFAIK it will fail

PR #444, opened 2024-06-13
Mount agent job socket/token

jradtilbrook:
If a user wants to mount the `buildkite-agent` to use it within the container, we should also automatically pass through the job API env vars, so that anything they might be wanting to do with the agent will just seamlessly work. Ie with this change, users will be able to use `buildkite-agent secret get` functionality as well.

I don't think its worth creating another plugin option for these. It makes sense to wrap up into mounting the agent as it applies to its usage


pzeballos (approved):
looks good 👍🏻

toote (changes requested):
While I believe it is a great implementation, I think that if the environment variable `BUILDKITE_AGENT_JOB_API_SOCKET` is not set (like a very old agent or the experiment being turned off) this would create a `-v ":" stanza that would be invalid so you should add the parameters inside an `if` statement that checks that the variable exists

jradtilbrook:
> While I believe it is a great implementation, I think that if the environment variable `BUILDKITE_AGENT_JOB_API_SOCKET` is not set (like a very old agent or the experiment being turned off) this would create a `-v ":"` stanza that would be invalid so you should add the parameters inside an `if` statement that checks that the variable exists (and - while you are at it - add a test to check the behaviour when it is not set)

Good catch! I've fixed that now

To test both paths, I opted to leave the v1 tests as is (without the env var exported) to confirm it works without that. And the current tests export the var to assert it is added to the CLI

toote (approved):
Thanks for the fixes... note that we will eventually deprecate the `v1` version and delete those tests so I'd still rather the latest tests check both scenarios

PR #448, opened 2024-07-10
Fix that allows error message to be displayed properly

karensawrey:
As per a support ticket.

Before this fix, the error message from line 81 was never shown because of `set -u`  on line 2.


tomowatt (approved):
🚀

pzeballos (approved):
They are correct :)

karensawrey:
Thanks for the approval! Will merge after running the test.

PR #449, opened 2024-07-11
Add option to build with `--with-dependencies`

mtollie:
In [compose >= v2.2.4.0](https://github.com/docker/compose/releases/tag/v2.24.0) there is a new option `--with-dependencies` which can be used to "Also build dependencies (transitively)" added in this PR: https://github.com/docker/compose/pull/11290.

Add support for passing this option to `docker compose build`.

Lots of copy/pasting so I'm not sure if I did this right. And let me know if more context would be helpful.


pzeballos (approved):
Thanks @mtollie! It looks good 👍🏻

Issue #457, opened 2024-09-16
Document Docker Image Caching changes with v5 release


tomowatt:
Previous versions of the Plugin, enabled pulling images that were specified via the `cache-from` parameter. This functionality is no longer available as of v5, due to the inclusion of being able to pass Images in different formats for the `cache-from` parameter.

Although the breaking change is documented, there is currently no publically available solution to keep this functionality as it was relied upon heavily for it's simplicity. There are workarounds available such as using `docker-container` driver and using the `cache_from` & `cache_to` docker-compose YAML attributes instead but these are not included anywhere in the repository.

This has lead to multiple instances of confusion from users of the Plugin, who upgraded to v5.x and found this loss of functionality and did not have enough information to move forward to get images pulled and to be used as cache.

Example workaround with docker-compose file:
```YAML
services:
  app:
    build:
    context: ../
    dockerfile: ./Dockerfile
    cache_to:
      - "type=registry,mode=max,image-manifest=true,oci-mediatypes=true,ref={Account ID}.dkr.ecr.us-east-1.amazonaws.com/{Cache Name}:cache"
```

and using native docker commands to push and use cache using `docker-container` driver:

```bash
$ docker buildx create --name container --driver=docker-container --use --bootstrap
$ docker buildx build --push -t {Account ID}.dkr.ecr.us-east-1.amazonaws.com/{Cache Name}:latest --cache-to type=registry,mode=max,image-manifest=true,oci-mediatypes=true,ref={Account ID}.dkr.ecr.us-east-1.amazonaws.com/{Cache Name}:cache --cache-from type=registry,ref={Account ID}.dkr.ecr.us-east-1.amazonaws.com/{Cache Name}:cache --cache-from type=registry,ref={Account ID}.dkr.ecr.us-east-1.amazonaws.com/{Cache Name}:main .
```

tomowatt:
By default, most Docker instances will use the `docker` build driver, but this by default does not support Cache Exports without changing the Docker Daemon itself via it's config. The `docker-container` driver is available with most Docker installs and can utilise Cache Exports and pulling from remote as a part of Image Building. So potential, the Plugin could include parameters and/or hooks to enable changing the Build Driver easily for uses.

With this the Plugin currently only has a `cache-from` parameter and not a `cache-to`, which should map to `cache_to` in the docker-compose file. This would allow more abstraction so Cache does not need to be specified from the docker-compose file.

[...]

PR #459, opened 2024-10-03
SUP-2585: Docker Cache Support

tomowatt:
Per Issue: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/457

There was changes from v4 to v5 which affected how Plugin manages Image Cache (prior to v5 it would pull images listed in `cache-from`) and now relies on the Docker Engine itself to pull/push image cache but now supports different cache formats e.g., OCI.

The Plugin currently does not have a `cache-to` parameter which is required for pushing image cache or changing the Docker Driver used to fully utilise the Docker Engine to pull image cache from registries.

### DIFF
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -73,7 +73,7 @@ for service_name in $(plugin_read_list BUILD) ; do
     [[ -n "${label:-}" ]] && labels+=("${label}")
   done <<< "$(plugin_read_list BUILD_LABELS)"
 
-  if [[ -n "${target}" ]] || [[ "${#labels[@]}" -gt 0 ]]|| [[ "${#cache_to[@]}" -gt 0 ]] || [[ "${#cache_from[@]}" -gt 0 ]]; then
+  if [[ -n "${target}" ]] || [[ "${#labels[@]}" -gt 0 ]] || [[ "${#cache_to[@]}" -gt 0 ]] || [[ "${#cache_from[@]}" -gt 0 ]]; then
     build_images+=("$service_name" "${image_name}" "${target}")
 
     build_images+=("${#cache_from[@]}")
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -35,8 +35,24 @@ get_caches_for_service() {
   fi
 }
 
+get_caches_to_service() {
+  local service="$1"
+
+  # Read any cache-to parameters provided
+  for line in $(plugin_read_list CACHE_TO) ; do
+    IFS=':' read -r -a tokens <<< "$line"
+    service_name=${tokens[0]}
+    service_image=$(IFS=':'; echo "${tokens[*]:1}")
+
+    if [ "${service_name}" == "${service}" ]; then
+      echo "$service_image"
+    fi
+  done
+}
+
+
 # Run through all images in the build property, either a single item or a list
-# and build up a list of service name, image name and optional cache-froms to
+# and build up a list of service name, image name and optional cache-froms and cache-tos to
 # write into a docker-compose override file
 for service_name in $(plugin_read_list BUILD) ; do
   target="$(plugin_read_config TARGET "")"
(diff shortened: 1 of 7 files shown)


commands/build.sh
```diff
@@ -47,19 +63,29 @@ for service_name in $(plugin_read_list BUILD) ; do
...
+    cache_to+=("$cache_line")
+  done
+
   labels=()
   while read -r label ; do
     [[ -n "${label:-}" ]] && labels+=("${label}")
   done <<< "$(plugin_read_list BUILD_LABELS)"
 
-  if [[ -n "${target}" ]] || [[ "${#labels[@]}" -gt 0 ]] || [[ "${#cache_from[@]}" -gt 0 ]]; then
+  if [[ -n "${target}" ]] || [[ "${#labels[@]}" -gt 0 ]]|| [[ "${#cache_to[@]}" -gt 0 ]] || [[ "${#cache_from[@]}" -gt 0 ]]; then
```
toote, line 76: just some spacing:

```suggestion
  if [[ -n "${target}" ]] || [[ "${#labels[@]}" -gt 0 ]] || [[ "${#cache_to[@]}" -gt 0 ]] || [[ "${#cache_from[@]}" -gt 0 ]]; then
```

commands/build.sh
```diff
@@ -35,8 +35,24 @@ get_caches_for_service() {
...
+  for line in $(plugin_read_list CACHE_TO) ; do
+    IFS=':' read -r -a tokens <<< "$line"
+    service_name=${tokens[0]}
+    service_image=$(IFS=':'; echo "${tokens[*]:1}")
+
+    if [ "${service_name}" == "${service}" ]; then
+      echo "$service_image"
+    fi
+  done
+}
```
toote, line 51: I believe it makes sense to have this new behaviour respect the `no-cache` option... and, as with the other it may make sense to unify it with the `get_caches_from_service` so that it has an additional parameter of what option to look for thus avoiding two functions that do exactly the same

plugin.yml
toote, line 1: it would be great if you rolled back all the whitespace changes in the file. While there's probably a linting/styling configuration that we need to standardize on - no matter what value that is nor our personal preferences - I would avoid it to make this PR easier to read as this kind of buries the fact that you added a new option

lib/shared.bash
```diff
@@ -220,6 +230,21 @@ function build_image_override_file_with_version() {
...
+        echo "https://docs.docker.com/compose/compose-file/compose-versioning/#versioning"
+        exit 1
+      fi
+
+      printf "      cache_to:\\n"
+      for cache_to_i in "${cache_to[@]}"; do
+        printf "        - %s\\n" "${cache_to_i}"
+      done
+    fi
+
```
toote, line 247: optional: may make sense to refactor this to its own function and use it twice instead of copy-pasting :P

plugin.yml
tomowatt: - [x] Whitespace changed reverted

commands/build.sh
tomowatt: @toote on fence on this one, as get where you are coming from but `--no-cache` is documented for not using cache _during_ the Building of an Image, where as `cache_to` is for specifying where cache should be exported to _after_ the Building of an Image. 

So feeling it wouldn't make sense to use `no-cache` to also block exporting cache, else if not wanting to export cache then `cache-to` values shouldn't be specified? 

https://docs.docker.com/reference/cli/docker/compose/build/
https://docs.docker.com/reference/compose-file/build/#cache_to

lib/shared.bash
tomowatt: - [x] refactored

commands/build.sh
tomowatt: though in retrospec it might make sense that `get_caches_for_service` is also not effected by `no-cache` parameter given itself configures the `--no-cache` flag itself for the `docker compose` commands which would ignore any `cache_from` values in the Docker Compose file

toote (approved):
:clap:

PR #468, opened 2025-01-28
feat: add support for GCP token propagation

valkum:
Hitting this now since we need credentials inside a container started with the docker-compose plugin.

This adds support for automatic GCP token propagation similar to the AWS token propagation.
Follows `docker-buildkite-plugin` see https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/269


pzeballos (approved):
thanks @valkum! This looks good. Can you also update the [README](https://github.com/valkum/docker-compose-buildkite-plugin/blob/gcp-token/docs/examples.md) with the next release version v5.7.0 (so the Linter doesn't complain)

PR #470, opened 2025-02-25
Update examples.md

melodicdan:
The latest version available is only `v5.6.0`

Updating docs to work when copy-pasted


pzeballos (commented):
Thanks @melodicdan! We actually need to do a new release, we were waiting if we can add one last thing before doing it, but this week we'll have 5.7 out :)

PR #463, opened 2024-10-28
SUP-2585: Add Support for changing Build Drivers

tomowatt:
Per Issue: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/457

To be able to fully support using the `cache-to` parameter, the Build Driver needs to be changed to use a different driver other than the default `docker` driver.
This PR will add the basic ability to change the driver used in a per Job basis and clean up the Build Instance afterwards to avoid issues of lingering Build Instances.

- [x] README.md updated
- [x] tests added
- [x] examples added to examples/examples.md

PR #477, opened 2025-04-11
Fix argument passing

jim-barber-he:
The builder.docker-opts and builder.platform arguments didn't work. The switch and its values were added to the builder_instance_args array as a single element, but the switch should be one element and its values the another. Otherwise it produces an invalid parameter when passed to the docker commands and errors are thrown.


nathanblogs (approved):

toote (approved):
I should have caught this when reviewing PR #463 :facepalm:

Thanks for fixing it!

PR #478, opened 2025-04-11
Change behavior of `builder.use` to only use builder once

KevinSnyderCodes:
When `builder.use` is `true`, this plugin calls `docker buildx use` which sets the default builder for the entire Buildkite agent.

This can cause issues for other Buildkite steps on the same agent. For example, another pipeline step might land on the same agent and expect to use the default `docker` driver, which does not require the user to specify `--load` when calling `docker build`. This results in the image not being available for subsequent commands (e.g. `docker tag` or `docker push`) and can break a pipeline that was previously working.

This change adds the `builder.use-once` parameter. When `true`, the plugin will pass the `--builder` flag to `docker build` instead of calling `docker buildx use`. This ensures that the specified builder is only used for this particular step.

---

We may want to recommend using `builder.use-once` over `builder.use` (or at least caution against using `builder.use`) due to the undesirable side effects of changing the default builder for the entire Buildkite agent.

If this is a change the team is open to accepting, I can work on adding tests and updating documentation.

### DIFF
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -111,6 +111,10 @@ fi
 
 build_params+=(build)
 
+if [[ -n "$(plugin_read_config BUILDER_NAME "")" ]] && [[ "$(plugin_read_config BUILDER_USE_ONCE "false")" == "true" ]]; then
+  build_params+=("--builder" "$(plugin_read_config BUILDER_NAME "")")
+fi
+
 if [[ ! "$(plugin_read_config SKIP_PULL "false")" == "true" ]] ; then
   build_params+=(--pull)
 fi
--- a/commands/build.sh
+++ b/commands/build.sh
@@ -111,12 +111,12 @@ fi
 
 build_params+=(build)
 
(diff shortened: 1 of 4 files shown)


noizwaves:
If the intended behavior of `use` is to use the builder for only the current job, perhaps the behavior introduced in this PR could be applied to `use` instead?

KevinSnyderCodes:
I introduced the `use-only` parameter as to not break existing behavior, but we can modify the behavior of the existing `use` parameter if preferred.

toote:
Given that the docker daemon is indeed an external global state shared by all agents that connect to it I agree that we should try to keep things as isolated as possible and the `builder.use` should indeed do what your `builder.use-once` now does - with the appropriate caveats added to the documentation.

Instead of adding a new option maybe the `use` option could be changed to be a ternary instead: `false` (current default), `true` (what you now describe as `use-once`) and something like `default` or `all` (which would be what now is `true`). That way we would still allow for the existing behaviour for those who really depend on it but they must really want to do so. What do you think?

KevinSnyderCodes:
@toote I've updated the behavior of `builder.use` to only use the builder once. We can revisit adding other options if they are requested, but I don't see any reason to add these options now.

toote (approved):

PR #480, opened 2025-04-23
Quiet push

### LINKED ISSUE #475: Pass --quiet to docker compose push
Pushing images is very verbose:

Image

I'd like to be able to forward the `--quiet` option to `docker compose push` (https://github.com/docker/compose/pull/9168).

There's already a `--quiet-pull` option, so maybe `--quiet-push` would suit 😃

toote:
Adds a new option to pass `--quiet` to push comands. While I was at it, I had the `quiet-pull` option be respected when the `pull` happens during a `push` stage.

Closes #475

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,9 +273,15 @@ Sets `docker-compose` to run with `--verbose`
 
 The default is `false`.
 
-#### `quiet-pull` (run only, boolean)
+#### `quiet-pull` (run/push only, boolean)
 
-Start up dependencies with `--quiet-pull` to prevent even more logs during that portion of the execution.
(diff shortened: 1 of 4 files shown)


azz:
Ah yes, this line does not respect `quiet-pull: true` from what I can tell?
https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/58ec2f9f512245bf914a1ae19ba266386f54e174/commands/run.sh#L84

toote:
@azz good catch! Added it to the code path

pzeballos (approved):

azz:
Looks great.

image

PR #491, opened 2025-05-08
Plugin Version Update

pzeballos:
v5.8.0 -> v5.9.0


pzeballos:
Updated in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/488

PR #493, opened 2025-05-22
Add option to allow env var usage in `push` strings

lparry:
Mirroring how it is done for `volume`, this PR adds a boolean config option, `expand-push-vars`, which enables runtime interpolation of variables within `push` configuration strings (and all the same risks of eval'ing strings).


toote (approved):
👏

PR #497, opened 2025-07-16
Store Docker digest in metadata

KevinGrandon:
This is something that we're currently using in order to leverage docker containers across pipelines. Possibly there's a better way of doing this? Aliases perhaps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,28 @@ Whether to set the metadata aboout the image for a service being pushed.
 
 Default: `true`.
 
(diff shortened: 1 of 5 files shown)


toote:
> I didn't see the prebuilt image URI with digest stored in metadata.

It is not, it is stored as the pushed tag. If you want or need the digest my original suggestion would still stand: a boolean flag to use the tag (current behavior) or the digest. What do you think?

PR #523, opened 2026-06-15
accept key value list and map format for env field

Damilola-obasa:
The env and environment fields previously only accepted a list format (`KEY=value)`, which differs from how env vars are defined at the pipeline and step level in Buildkite.This change makes the plugin accept both the list format and the map format (KEY: value), so users can write env vars in whichever way they prefer

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -114,6 +114,7 @@ fi
 
 # append env vars provided in ENV or ENVIRONMENT, these are newline delimited
 # supports both list format (- KEY=value) and map format (KEY: value)
+# PROPAGATION_LIST excludes env-propagation-list, which shares the ENV_ prefix when flattened
 while IFS=$'\n' read -r env ; do
   [[ -n "${env:-}" ]] && run_params+=("-e" "${env}")
 done <<< "$(printf '%s\n%s' \
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -113,11 +113,12 @@ if [[ "$(plugin_read_config RUN_LABELS "true")" =~ ^(true|on|1)$ ]]; then
 fi
 
 # append env vars provided in ENV or ENVIRONMENT, these are newline delimited
+# supports both list format (- KEY=value) and map format (KEY: value)
 while IFS=$'\n' read -r env ; do
   [[ -n "${env:-}" ]] && run_params+=("-e" "${env}")
 done <<< "$(printf '%s\n%s' \
-  "$(plugin_read_list ENV)" \
-  "$(plugin_read_list ENVIRONMENT)")"
+  "$(plugin_read_list_or_map ENV)" \
+  "$(plugin_read_list_or_map ENVIRONMENT)")"
 
 # Propagate all environment variables into the container if requested
 if [[ "$(plugin_read_config PROPAGATE_ENVIRONMENT "false")" =~ ^(true|on|1)$ ]] ; then
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -117,7 +117,7 @@ fi
 while IFS=$'\n' read -r env ; do
   [[ -n "${env:-}" ]] && run_params+=("-e" "${env}")
 done <<< "$(printf '%s\n%s' \
-  "$(plugin_read_list_or_map ENV)" \
+  "$(plugin_read_list_or_map ENV PROPAGATION_LIST)" \
   "$(plugin_read_list_or_map ENVIRONMENT)")"
 
 # Propagate all environment variables into the container if requested
--- a/lib/shared.bash
+++ b/lib/shared.bash
@@ -65,6 +65,24 @@ function prefix_read_list() {
   fi
 }
 
+# Reads either a value, a list, or a map from plugin config
+# For map format (YAML object), outputs KEY=value pairs
+function plugin_read_list_or_map() {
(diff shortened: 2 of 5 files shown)


plugin.yml
```diff
@@ -91,10 +91,10 @@ configuration:
     entrypoint:
       type: string
     env:
-      type: [ string, array ]
+      type: [ string, array, object ]
       minimum: 1
     environment:
-      type: [ string, array ]
+      type: [ string, array, object ]
```
petetomasik, line 97: `env`/`environment` now accept `object`, but the README still describes them as "string or array" / "A list of either KEY or KEY=VALUE" (README.md:78-90). Can we update the docs in this PR so the map form is discoverable?

lib/shared.bash
```diff
@@ -65,6 +65,29 @@ function prefix_read_list() {
...
+
+  if [[ -n "${!list_param:-}" ]] || [[ -n "${!prefix:-}" ]]; then
+    prefix_read_list "$prefix"
+  else
+    local env_prefix="${prefix}_"
+    local key
+    while IFS= read -r varname ; do
+      key="${varname#"${env_prefix}"}"
+      if ! in_array "${key}" "$@"; then
+        echo "${key}=${!varname}"
```
petetomasik, line 85: Edge case worth noting, probably document-only rather than a code change: when a map key flattens to a name with consecutive underscores, the agent also sets a collapsed-underscore deprecation alias. So `env: {MY__VAR: value}` produces both `BUILDKITE_PLUGIN_DOCKER_COMPOSE_ENV_MY__VAR` and `..._ENV_MY_VAR`, and this scan picks up both, passing `-e MY__VAR=value -e MY_VAR=value` to the container.

The list form never hit this, since the agent doesn't alias array element values. De-duping here isn't really feasible (the plugin can't distinguish a legitimately-collapsed key from an alias), and the agent already warns about the deprecated name, so I'd lean toward calling out the limitation rather than handling it in code. Flagging in case you'd rather handle it differently.

tests/run.bats
```diff
@@ -222,6 +222,54 @@ cmd3"
   unstub buildkite-agent
 }
 
+@test "Run without a prebuilt image with env as a map" {
```
petetomasik, line 225: One test-coverage gap worth filling while you're adding these: an `env` map combined with `env-propagation-list`. The `PROPAGATION_LIST` exclusion arg in `plugin_read_list_or_map ENV PROPAGATION_LIST` is the most fragile part of this change, and nothing exercises an env map alongside `env-propagation-list`. (The existing "Run with a list of propagated env vars" test guards against removing the exclusion arg, but not the co-occurrence.) A test that sets `ENV_PROPAGATION_LIST` plus the map keys and asserts both the propagated vars and the map vars appear, with no literal `-e PROPAGATION_LIST=...`, would lock it in.

lib/shared.bash
```diff
@@ -65,6 +65,29 @@ function prefix_read_list() {
...
+# Reads either a value, a list, or a map from plugin config
+# For map format (YAML object), outputs KEY=value pairs
+# Additional arguments are sibling config key suffixes to exclude from map scanning
+function plugin_read_list_or_map() {
+  local config_key="$1"
+  shift
+  local prefix="BUILDKITE_PLUGIN_DOCKER_COMPOSE_${config_key}"
+  local list_param="${prefix}_0"
+
+  if [[ -n "${!list_param:-}" ]] || [[ -n "${!prefix:-}" ]]; then
```
petetomasik, line 77: Minor, non-blocking: this condition re-implements the same list-vs-scalar check `prefix_read_list` already does internally (it tests `${prefix}_0`, then `${prefix}`, and emits nothing if neither is set). It's duplicated here only to decide whether to fall through to the map scan. Not wrong, and arguably clearer about intent than the alternatives. If you wanted to avoid the duplicated test, you could capture `prefix_read_list`'s output and map-scan only when it comes back empty, but that trades the duplicated check for a subshell capture.

lib/shared.bash
```diff
@@ -65,6 +65,29 @@ function prefix_read_list() {
...
+    prefix_read_list "$prefix"
+  else
+    local env_prefix="${prefix}_"
```
petetomasik: Nit: `env_prefix` reads as env-specific, but this function is generic (it's called for both `ENV` and `ENVIRONMENT`, and could serve any map-typed field). Something neutral like `scan_prefix` would match the rest of the file, which uses plain `prefix` for the analogous value.

commands/run.sh
```diff
@@ -113,11 +113,12 @@ if [[ "$(plugin_read_config RUN_LABELS "true")" =~ ^(true|on|1)$ ]]; then
 fi
 
 # append env vars provided in ENV or ENVIRONMENT, these are newline delimited
+# supports both list format (- KEY=value) and map format (KEY: value)
```
petetomasik, line 116: Tiny nit: this comment describes the YAML source shapes (`- KEY=value` / `KEY: value`), whereas the code right here works on already-flattened env vars. It's arguably clearer to a user framed this way, so I wouldn't change it; flagging only because the `- KEY=value` reads slightly oddly mid-script.

petetomasik (commented):
Some thoughts, nothing major or blocking beyond an update to the README.

petetomasik (approved):
✅

PR #526, opened 2026-07-09
fix: add timeouts to ensure_stopped and compose_cleanup to prevent orphaned containers on cancellation

Damilola-obasa:
Added timeout 30 to docker stop and timeout 10 to docker logs in ensure_stopped() so the trap handler always returns within a bounded time, leaving room for compose_cleanup() to run.
Also added --timeout 30 to docker compose stop (graceful shutdown path) and docker compose down in compose_cleanup() to guard against the same unbounded hang pattern there.

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -478,9 +478,9 @@ fi
 
 ensure_stopped() {
   echo '+++ :warning: Signal received, stopping container'
-  timeout 30 docker stop "${container_name}" || true
+  run_with_deadline 30 docker stop "${container_name}" || true
   echo '~~~ Last log lines that may be missing above (if container was not already removed)'
-  timeout 10 docker logs "${container_name}" || true
+  run_with_deadline 10 docker logs "${container_name}" || true
   exitcode='TRAP'
 }
 
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -478,9 +478,7 @@ fi
 
 ensure_stopped() {
   echo '+++ :warning: Signal received, stopping container'
-  run_with_deadline 8 docker stop "${container_name}" || true
-  echo '~~~ Last log lines that may be missing above (if container was not already removed)'
-  run_with_deadline 2 docker logs "${container_name}" || true
+  run_with_deadline 2 docker stop "${container_name}" || true
   exitcode='TRAP'
 }
 
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -478,9 +478,9 @@ fi
 
 ensure_stopped() {
   echo '+++ :warning: Signal received, stopping container'
-  docker stop "${container_name}" || true
+  timeout 30 docker stop "${container_name}" || true
   echo '~~~ Last log lines that may be missing above (if container was not already removed)'
-  docker logs "${container_name}" || true
+  timeout 10 docker logs "${container_name}" || true
   exitcode='TRAP'
 }
 
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -478,9 +478,9 @@ fi
 
 ensure_stopped() {
   echo '+++ :warning: Signal received, stopping container'
-  run_with_deadline 30 docker stop "${container_name}" || true
+  run_with_deadline 8 docker stop "${container_name}" || true
   echo '~~~ Last log lines that may be missing above (if container was not already removed)'
-  run_with_deadline 10 docker logs "${container_name}" || true
+  run_with_deadline 2 docker logs "${container_name}" || true
   exitcode='TRAP'
 }
 
--- a/lib/run.bash
+++ b/lib/run.bash
@@ -2,15 +2,16 @@
 
 compose_cleanup() {
   local FAILURES=0
+  local deadline="${_CLEANUP_DEADLINE:-30}"
 
   if [[ "$(plugin_read_config GRACEFUL_SHUTDOWN 'false')" == "false" ]]; then
-    SIGNAL_PARAMS=(kill)
+    SIGNAL="kill"
   else
-    SIGNAL_PARAMS=(stop --timeout 30)
+    SIGNAL="stop"
   fi
 
   # Send all containers the corresponding signal
-  if ! run_docker_compose "${SIGNAL_PARAMS[@]}"; then
+  if ! run_with_deadline "$deadline" run_docker_compose "${SIGNAL}"; then
     FAILURES=$((FAILURES + 1))
   fi
 
@@ -20,17 +21,17 @@ compose_cleanup() {
     RM_PARAMS+=(-v)
   fi
 
-  if ! run_docker_compose "${RM_PARAMS[@]}"; then
+  if ! run_with_deadline "$deadline" run_docker_compose "${RM_PARAMS[@]}"; then
     FAILURES=$((FAILURES + 1))
   fi
 
   # Stop and remove all the linked services and network
-  DOWN_PARAMS=(down --remove-orphans --timeout 30)
+  DOWN_PARAMS=(down --remove-orphans)
   if [[ "$(plugin_read_config LEAVE_VOLUMES 'false')" == "false" ]]; then
     DOWN_PARAMS+=(--volumes)
   fi
 
-  if ! run_docker_compose "${DOWN_PARAMS[@]}"; then
+  if ! run_with_deadline "$deadline" run_docker_compose "${DOWN_PARAMS[@]}"; then
     FAILURES=$((FAILURES + 1))
   fi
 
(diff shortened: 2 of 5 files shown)


commands/run.sh
```diff
@@ -478,9 +478,9 @@ fi
 
 ensure_stopped() {
   echo '+++ :warning: Signal received, stopping container'
-  docker stop "${container_name}" || true
+  timeout 30 docker stop "${container_name}" || true
```
scadu: timeout is not available by default on macOS, and `|| true` hides that failure, so neither Docker command would run. Also, these operations can take 40 seconds while Buildkite’s default cancellation grace period is 10 seconds. Could we use a portable deadline shorter than the cancellation window? That'd most probably be mix of `sleep` and `kill` 🤔

lib/run.bash
```diff
@@ -4,13 +4,13 @@ compose_cleanup() {
   local FAILURES=0
 
   if [[ "$(plugin_read_config GRACEFUL_SHUTDOWN 'false')" == "false" ]]; then
-    SIGNAL="kill"
+    SIGNAL_PARAMS=(kill)
   else
-    SIGNAL="stop"
```
scadu, line 9: Compose’s `--timeout` is a container shutdown grace period, not a wall-clock command timeout. These commands can still hang, and the change increases the usual grace period from 10 to 30 seconds. Could we instead bound each cleanup command independently so later cleanup still runs after a timeout?

tests/docker-compose-cleanup.bats
```diff
@@ -15,31 +15,31 @@ setup () {
   stub stubbed_run_docker_compose \
     "kill : echo \$@" \
```
scadu, line 44: These tests only verify arguments, so they do not catch a hanging command or a missing timeout executable. Could we add a behavioral test confirming cleanup finishes within its deadline and continues to later operations?

commands/run.sh
```diff
@@ -478,9 +478,9 @@ fi
 
 ensure_stopped() {
   echo '+++ :warning: Signal received, stopping container'
-  docker stop "${container_name}" || true
+  run_with_deadline 30 docker stop "${container_name}" || true
   echo '~~~ Last log lines that may be missing above (if container was not already removed)'
-  docker logs "${container_name}" || true
+  run_with_deadline 10 docker logs "${container_name}" || true
```
scadu: These sequential deadlines still allow 40 seconds, while Buildkite’s default cancellation grace period is 10 seconds. The bootstrap may therefore be killed before pre-exit cleanup runs. Could we use an aggregate budget below 10 seconds, prioritizing stop and tightly limiting or skipping logs?

tests/docker-compose-cleanup.bats
```diff
@@ -11,6 +11,17 @@ setup () {
...
+@test "run_with_deadline allows a fast command to complete" {
+  run run_with_deadline 5 echo "done"
+  assert_success
+  assert_output "done"
+}
+
+@test "run_with_deadline terminates a command that exceeds the deadline" {
+  run run_with_deadline 1 sleep 100
+  assert_failure
+}
```
scadu, line 31: These tests only cover commands that exit normally or respond to SIGTERM, so they pass even though a resistant command or descendant can outlive the deadline. Could we assert elapsed time and verify that a TERM-resistant process tree and the watchdog are both gone afterward?

lib/shared.bash
```diff
@@ -35,6 +35,21 @@ function plugin_prompt_and_must_run() {
...
+  "$@" &
+  local cmd_pid=$!
+  ( sleep "$seconds" && kill "$cmd_pid" 2>/dev/null ) &
+  local watcher_pid=$!
+  wait "$cmd_pid" 2>/dev/null
+  local status=$?
+  kill "$watcher_pid" 2>/dev/null
+  wait "$watcher_pid" 2>/dev/null || true
+  return "$status"
+}
```
scadu: At the moment, this only signals the immediate PID, so a wrapped command can leave descendants running. Could we isolate each background job into a process group and escalate TERM to KILL?

```suggestion
function run_with_deadline() (
  local timeout="$1"
  shift

  # Give each background job its own process group.
  set -m
  "$@" &
  local command_pid=$!

  (
    sleep "$timeout"
    kill -TERM -- "-$command_pid" 2>/dev/null || exit 0
    sleep 1
    kill -KILL -- "-$command_pid" 2>/dev/null || true
  ) &
  local watchdog_pid=$!

  wait "$command_pid" 2>/dev/null
  local status=$?

  if kill -0 -- "-$command_pid" 2>/dev/null; then
    wait "$watchdog_pid" 2>/dev/null || true
  else
    kill -TERM -- "-$watchdog_pid" 2>/dev/null || true
    wait "$watchdog_pid" 2>/dev/null || true
  fi

  return "$status"
)
```

scadu (commented):
A few more more things I spotted. Most important is the `run_with_deadline` function. Otherwise looking good ;)

commands/run.sh
Damilola-obasa: Yes that very true. Made this adjustment

lib/shared.bash
```diff
@@ -35,6 +35,37 @@ function plugin_prompt_and_must_run() {
...
+
+  if kill -0 -- "-$cmd_pid" 2>/dev/null; then
+    wait "$watcher_pid" 2>/dev/null || true
+  else
+    kill -TERM -- "-$watcher_pid" 2>/dev/null || true
+    wait "$watcher_pid" 2>/dev/null || true
+  fi
+
+  return "$status"
+)
```
scadu, line 67: Looks like there's still an [off-by-one](https://en.wikipedia.org/wiki/Off-by-one_error) here that I've missed earlier.
We wait N seconds before sending `TERM`, then another second before `KILL`, so a two-second deadline can actually take three seconds.

That extra second matters because `pre-exit` runs in the same cancellation window. Could we send `TERM` one second earlier and keep `KILL` at the requested deadline?

```suggestion
function run_with_deadline() (
     local seconds="$1"; shift
     local term_after=$((seconds > 1 ? seconds - 1 : 0))

     set -m
     "$@" &
     local cmd_pid=$!

     (
       sleep "$term_after"
       kill -TERM -- "-$cmd_pid" 2>/dev/null || exit 0
       sleep "$((seconds - term_after))"
       kill -KILL -- "-$cmd_pid" 2>/dev/null || true
     ) &
     local watcher_pid=$!

     wait "$cmd_pid" 2>/dev/null
     local status=$?

     if kill -0 -- "-$cmd_pid" 2>/dev/null; then
       wait "$watcher_pid" 2>/dev/null || true
     else
       kill -TERM -- "-$watcher_pid" 2>/dev/null || true
       wait "$watcher_pid" 2>/dev/null || true
     fi

     return "$status"
   )
```

commands/run.sh
```diff
@@ -478,9 +478,9 @@ fi
 
 ensure_stopped() {
```
scadu, line 479: To consider: could we limit `docker stop` to two seconds and skip collecting logs here?

Right now this handler can spend eight seconds stopping the run container and another two reading its logs. That can use the agent's entire default cancellation window before `pre-exit` gets a chance to clean up the rest of the Compose project, which is the problem reported in #517.

Two seconds still lets Docker handle the quick path, but stops a stuck call from blocking `pre-exit`. I'd skip `docker logs` because it's another Docker call that can hang and doesn’t help remove any resources. The existing 30-second limits in `compose_cleanup` can stay as they are.

tests/docker-compose-cleanup.bats
```diff
@@ -11,6 +11,34 @@ setup () {
...
+
+@test "run_with_deadline kills TERM-resistant processes via SIGKILL escalation" {
+  local start=$SECONDS
+  # bash ignores SIGTERM; only SIGKILL from the escalation path can stop it
+  run run_with_deadline 2 bash -c 'trap "" TERM; while true; do sleep 9998; done'
+  assert_failure
+  [[ $((SECONDS - start)) -lt 6 ]]
+  ! pgrep -f "sleep 9998" >/dev/null 2>&1
+}
+
```
scadu, line 41: If we adjust `run_with_deadline` as above https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/526/changes#r3734388433, could we add a regression test for the split? The existing tests cover process cleanup, but their timing assertions don't catch the extra
 second.
```suggestion
@test "run_with_deadline includes escalation in the deadline" {
     local sleep_calls="${BATS_TEST_TMPDIR}/sleep-calls"

     sleep() {
       printf '%s\n' "$1" >> "$sleep_calls"
     }

     run run_with_deadline 5 bash -c 'trap "" TERM; while :; do :; done'
     assert_failure

     run cat "$sleep_calls"
     assert_success
     assert_output $'4\n1'
   }
```

scadu (approved):

scadu:
@Damilola-obasa good stuff! Feel free to merge it 🚀

PR #527, opened 2026-07-17
Add `bake` option to build and push with `docker buildx bake`

olihou:
## Summary

Adds an opt-in, build-only `bake` boolean that builds services with `docker buildx bake --push` instead of `docker compose build`.

With the `docker-container` and `remote` build drivers, the built image lives in BuildKit's own store, so `docker compose build` exports it as a tarball and imports it into the Docker daemon before it can be pushed. For large images that load dominates the step **even on a full cache hit** — Windows containers are the extreme case, with multi-gigabyte base layers.

`bake --push` uses BuildKit's registry exporter, so the image goes straight from BuildKit to the registry and the daemon is never involved. In our CI (Windows native builds on a `remote` builder) this took the image build+push step from ~12 min — almost entirely tarball export/import, everything else cached — down to ~1–2 min.

## Behaviour

- Opt-in; default `false`, so existing behaviour is unchanged.
- Reuses the generated override file and Compose config files, so `image`, `cache-from`, `cache-to`, `target` and `build-labels` all apply (bake reads them via `--file`).
- Honours `builder`, `no-cache`, `skip-pull`, `ssh`, `buildkit-inline-cache` and `args`.
- Because it pushes, the pushed image is recorded in the build metadata (unless `push-metadata: false`), so later `run`/`push` steps use it — exactly like a regular `push`.

```yml
steps:
  - plugins:
      - docker-compose#v5.13.0:
          build: app
          bake: true
          builder:
            name: container
            use: true
            create: true
            driver: docker-container
```

## Notes / limitations

- The service must define an `image` with a registry reference, and the agent must be authenticated for that registry.
- bake pushes during the build, so a separate `push` entry for the same service isn't needed.
- Requires BuildKit (no effect under `cli-version: 1` unless `buildkit` is also enabled).

## Testing

- `shellcheck` clean.
- Added unit tests in `tests/build.bats` (build + push + metadata; builder/no-cache/inline-cache/args; cache-from override file). The full `bats tests` suite passes.
- `plugin-linter` passes (plugin.yml valid; all README/examples config examples valid).

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,22 @@ Assuming you have a compatible Docker installation and configuration in the agen
 
 You may want to also add `BUILDKIT_INLINE_CACHE=1` to your build arguments (`args` option in this plugin), but know that [there are known issues with it](https://github.com/moby/buildkit/issues/2274).
 
+#### `bake` (build only, boolean)
+
+Build the services with [`docker buildx bake`](https://docs.docker.com/build/bake/) instead of `docker compose build`, pushing the resulting image(s) straight to the registry (`bake --push`) rather than loading them into the local Docker daemon.
+
+This is useful with the `docker-container` and `remote` build [drivers](https://docs.docker.com/build/builders/drivers/): with those drivers the built image lives in BuildKit's own store, so `docker compose build` must export it as a tarball and import it into the daemon before it can be pushed. For large images — Windows containers are the extreme case, with multi-gigabyte base layers — that load dominates the step even on a full cache hit. Pushing directly from BuildKit skips it entirely.
+
+The image tags, `cache-from`, `cache-to`, `target` and `build-labels` options are honoured (bake reads them from the Compose config and the generated override file), as are `builder`, `no-cache`, `skip-pull`, `ssh`, `buildkit-inline-cache` and `args`. Because the image is pushed rather than loaded, the pushed image is recorded in the build metadata (unless `push-metadata` is `false`) so later `run` and `push` steps use it, just like a regular `push`.
+
+Notes:
+
+- The service must define an `image` with a registry reference to push to, and the agent must be authenticated for that registry.
+- As bake pushes during the build, you do not need a separate `push` entry for the same service in this step.
+- Requires BuildKit; it has no effect with `cli-version: 1` unless `buildkit` is also enabled.
+
+Default: `false`
+
 #### `ssh` (build only, boolean or string)
 
 It will add the `--ssh` option to the build command with the passed value (if `true` it will use `default`). Note that it assumes you have a compatible Docker installation and configuration in the agent (meaning you are using BuildKit and it is correctly setup).
--- a/lib/bake.bash
+++ b/lib/bake.bash
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+BAKE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# compose_image_for_service lives in push.bash; source it so the pushed image can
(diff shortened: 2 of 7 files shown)


lib/bake.bash
```diff
@@ -0,0 +1,97 @@
...
+# and import it into the daemon before it can be pushed. For large images (for
+# example Windows containers, whose base layers are several GB) that load
+# dominates the step even on a full cache hit. `bake --push` uses BuildKit's
+# registry exporter, so the image goes from BuildKit to the registry directly and
+# the daemon is never involved.
+#
+# The generated override file and the Compose config files carry the image tags,
+# cache_from/cache_to, target and labels, and bake reads them through `--file`,
+# so those options apply unchanged.
+function build_with_bake() {
```
Damilola-obasa: Thanks a lot! was reviewing this and noticed `build_with_bake` doesn't check `with-dependencies` the way the regular build path does. If someone sets `with-dependencies: true` expecting deps to build first, this could just quietly skip that.

lib/bake.bash
```diff
@@ -0,0 +1,97 @@
...
+  local services=("$@")
+
+  local override_file="docker-compose.buildkite-${BUILDKITE_BUILD_NUMBER}-override.yml"
```
Damilola-obasa: One small thing here, `override_file `and `group_type` get set up here, but build.sh already worked out the exact same values right before calling this function. Nothing too major or broken here, just a bit of duplicated logic that could drift out of sync if either side changes later. Thinking this might be worth passing as arguments instead at some point

lib/bake.bash
```diff
@@ -0,0 +1,97 @@
...
+  echo "${group_type} :docker: Building and pushing services with bake: ${services[*]}"
+  plugin_prompt_and_must_run docker "${bake_params[@]}"
+
+  # Record the pushed image for each service so later run/push steps pull it
+  # instead of falling back to a rebuild (mirrors the push command's behaviour).
+  if [[ "$(plugin_read_config PUSH_METADATA "true")" == "true" ]] ; then
+    local prebuilt_image_namespace service image
+    prebuilt_image_namespace="$(plugin_read_config PREBUILT_IMAGE_NAMESPACE 'docker-compose-plugin-')"
+    for service in "${services[@]}" ; do
+      image="$(compose_image_for_service "$service")"
```
Damilola-obasa: Not a major blocker here but this ends up running docker compose config once for every service, to grab its image name. Probably not noticeable with a handful of services, but if building a long list it could start to add up. It may be better to grab the config once and reusing it instead

README.md
```diff
@@ -396,6 +396,22 @@ Assuming you have a compatible Docker installation and configuration in the agen
...
+
+The image tags, `cache-from`, `cache-to`, `target` and `build-labels` options are honoured (bake reads them from the Compose config and the generated override file), as are `builder`, `no-cache`, `skip-pull`, `ssh`, `buildkit-inline-cache` and `args`. Because the image is pushed rather than loaded, the pushed image is recorded in the build metadata (unless `push-metadata` is `false`) so later `run` and `push` steps use it, just like a regular `push`.
+
+Notes:
+
+- The service must define an `image` with a registry reference to push to, and the agent must be authenticated for that registry.
+- As bake pushes during the build, you do not need a separate `push` entry for the same service in this step.
+- Requires BuildKit; it has no effect with `cli-version: 1` unless `buildkit` is also enabled.
```
Damilola-obasa: This note says bake has no effect under `cli-version: 1` without buildkit enabled, but looking at `build.sh` it looks like `build_with_bake` gets called regardless of `cli-version/buildkit`.

